PR #27, opened 2015-01-31
add gist url in manifest

petereberlecom:
add gist url in manifest


stefanbuck:
Cool, will merge it with the upcoming core release https://github.com/github-linker/core/milestones/1.7.0

PR #59, opened 2015-12-31
Update README.md

teokitan:
I think Copyright 2015 is needed


stefanbuck:
Thanks for the hint, but maybe "(c) Stefan Buck" is enough – I have to do some research on this.

timdorr:
Facebook does "start year-present" in their licenses: https://github.com/facebook/react/blob/master/LICENSE

So, "(c) 2014-present" should be sufficient.

stefanbuck:
Thanks @teokitan and @timdorr. I've updated the copyright https://github.com/octo-linker/chrome-extension/pull/64

PR #76, opened 2016-05-16
Fix intra-repo links in DEVELOPER.md

josephfrazier:
Before, they were leading me to 404s like https://github.com/OctoLinker/browser-extension/blob/master/blob/master/lib/plugins (note the extra "/blob/master")


stefanbuck:
:+1:

PR #80, opened 2016-05-19
Allow for exported imports.

### LINKED ISSUE #78: Can't resolve imports of a directory
Assuming the following directory structure:

```
index.js
foo/
  index.js
```

And in the root `index.js`, you've got:

```js
import foo from './foo';
```

It is unable to resolve it to `./foo/index.js`

jamestalmage:
The module spec allows for the following:

```js
export * from "foo";
export {foo, bar} from "foo";
export {foo as bar} from "foo";
```

Reference:
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#Syntax
 - http://www.ecma-international.org/ecma-262/6.0/#sec-exports

Fixes #78


stefanbuck:
Looks good 👍  Will publish a new version this evening. Thank you

stefanbuck:
https://github.com/OctoLinker/browser-extension/releases/tag/v4.0.2 🎉 
Will be available in the next minutes in the chrome store.

PR #85, opened 2016-05-26
Changed the notification link to open in a new tab

dpeukert:


stefanbuck:
Hi @dpeukert, mhhh link handling is always tricky, everyone wants to have it differently 😄  I used a "normal" link so you can force a new tab by CMD+Click or with the context menu or just click it to open the link in the same tab. If the link would be `_blank` by default this wouldn't be possible anymore. 

Convinced?

RickvdStaaij:
Personally I think this is a case where you do not want to disturb the flow that a user is currently in. I was like "what was I doing again?" after I clicked the link. You are offering some extended info on a page where the user is for a reason, different from checking your update message. So I think in this case it makes more sense to have a target _blank in there.

Just my 2 cents :wink:

stefanbuck:
I'm happy to merge this PR if that is the desire of the majority 😉 
Btw. do you find the notification useful or should I communicate new releases only via twitter?

RickvdStaaij:
I do think it's a nice addition!

dpeukert:
@RickvdStaaij That was my thinking as well.

stefanbuck:
Thank you guys. Will be part of the next release.

Issue #49, opened 2015-08-29
Opening in new tab


ewnd9:
Hello, thanks for the great extension. What about opening package in new tab on middle mouse click?

stefanbuck:
Should work since `v3.8.3`

ewnd9:
As I found in github-linker/chrome-extension#47 there is also `ctrl+click` which works for me on `v3.8.3`, but not middle mouse click.

stefanbuck:
What is your OS?

ewnd9:
Ubuntu 14.04

stefanbuck:
I don't have neither Ubuntu or a three button mouse to test it. So could you please got to [this](http://output.jsbin.com/jorejatiwa/) page and press the button with the middle mouse and reply the result here. Thank you

ewnd9:
Event value: 2

stefanbuck:
Mhhh it should work, because it's already implemented like this. Can you please test if the middle mouse click works for this demo http://output.jsbin.com/yowofoxisu/ ?

ewnd9:
Yep, it works in demo

stefanbuck:
It's nearly the same code. I`ll looking for an ubuntu image and a test mouse.

williamboman:
It would also be nice if the `github-link` span element was an anchor element with an href that could be opened in a new context. Simply like this;
![eh](https://cloud.githubusercontent.com/assets/6705160/9640016/70227ea2-51af-11e5-8195-fa53cdb1e80d.gif)

This could probably be achieved through a HTML/JS data-URI or simply resolve every dependency in a batch request and insert the links in the DOM when it resolves (didn't it work like this before?).

stefanbuck:
Yes, in the previous version it worked like that, because the extension had a huge list of all repo urls. unfortunately this cause a big performance issue #38. So I decided to replace that static url lists with an api service https://github.com/github-linker/live-resolver/. That's the reason why the behaviour is now a little bit different. Solving all links by default would be overkill and to heavy for the API / my heroku plan. 

I will consider / think about a solution for the next major release 4.0.

williamboman:
With my limited knowledge of how the extension works, wouldn't it be possible to simply href all dependencies to e.g. https://githublinker.herokuapp.com/npm/foo-bar?redirect=1, with a `?redirect=1` flag which tells the server to respond with a 301 (perhaps 302)?

[...]

PR #87, opened 2016-05-27
Use <a href> instead of <span>, with 302 redirection for liveResolverQuery urls

josephfrazier:
This depends on https://github.com/OctoLinker/live-resolver/pull/12 being merged.

As discussed in https://github.com/OctoLinker/browser-extension/issues/49#issuecomment-137948995.

This is a proof-of-concept and may need to be refined to handle error/edge cases.

### DIFF
--- a/lib/insert-link.js
+++ b/lib/insert-link.js
@@ -4,7 +4,7 @@ import findAndReplaceDOMText from 'findandreplacedomtext';
 const CLASS_NAME = 'octo-linker-link';
 
 function createLinkElement(text, dataAttr = {}) {
-  const linkEl = document.createElement('span');
+  const linkEl = document.createElement('a');
 
   // Set link text
   linkEl.innerHTML = text;
--- a/lib/insert-link.js
+++ b/lib/insert-link.js
@@ -1,5 +1,6 @@
 import $ from 'jquery';
 import findAndReplaceDOMText from 'findandreplacedomtext';
+import liveResolverQuery from './resolver/live-resolver-query.js';
 
 const CLASS_NAME = 'octo-linker-link';
 
@@ -17,6 +18,12 @@ function createLinkElement(text, dataAttr = {}) {
     if (dataAttr.hasOwnProperty(key)) {
       linkEl.dataset[key] = dataAttr[key];
     }
+  }
+
+  if (dataAttr.resolver == 'liveResolverQuery') {
+    const type = dataAttr.type;
+    const target = dataAttr.target;
+    linkEl.href = liveResolverQuery({type, target}, true).url;
   }
 
   return linkEl;
--- a/lib/click-handler.js
+++ b/lib/click-handler.js
@@ -58,6 +58,8 @@ function onClick(event) {
     const newWindow = (event.metaKey || event.ctrlKey || event.which === 2);
     openUrl((res || {}).url || url, newWindow);
   });
+
+  return false;
 }
 
 export default function () {
--- a/lib/resolver/live-resolver-query.js
+++ b/lib/resolver/live-resolver-query.js
@@ -1,6 +1,6 @@
-export default function ({ type, target }) {
+export default function ({ type, target }, redirect=false) {
   return {
-    url: `https://githublinker.herokuapp.com/q/${type}/${target}`,
+    url: `https://githublinker.herokuapp.com/q/${type}/${target}` + (redirect ? '?redirect=1' : ''),
     method: 'GET',
   };
 }
--- a/test/insert-link.test.js
+++ b/test/insert-link.test.js
@@ -19,9 +19,9 @@ describe('helper-replace-keywords', () => {
       dataAttributes += ` data-${key}="${value}"`;
     }
 
-    const start = `<span class="octo-linker-link"${dataAttributes}>`;
+    const start = `<a class="octo-linker-link"${dataAttributes}>`;
     const input = el.replace(/\$0/g, '');
-    const output = el.replace('$0', start).replace('$0', '</span>');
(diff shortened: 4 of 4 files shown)


lib/insert-link.js
```diff
@@ -19,6 +20,12 @@ function createLinkElement(text, dataAttr = {}) {
     }
   }
 
+  if (dataAttr.resolver == 'liveResolverQuery') {
+    const type = dataAttr.type;
+    const target = dataAttr.target;
+    linkEl.href = liveResolverQuery({type, target}, true).url;
+  }
```
stefanbuck: In same cases the liveResolver have to be called multiple times for the same target. For example in a JavaScript file like `require('foo')` the target `foo` could be either a bower or npm module.

Maybe the [live-resolver](https://github.com/OctoLinker/live-resolver) should accept multiple types like `https://githublinker.herokuapp.com/q/?type=npm,bower&target=foo`

stefanbuck:
First thanks for picking up this issue again. Ideally every OctoLinker link should be an anchor element so it will behave like every other link on the web. This PR will changes this for some of them and other will stay the same and here is where I see a problem. Imagine the follow scenario. A JavaScript file with two import statements where the first one imports jquery and the second one loads a local file. Our example user prefers to open all links in a new tab with the content menu (right click > Open link in new tab). This would work for the first link, because this is an anchor element. But for the second one, it won't work like the previous one, because it still just a span element. 

I'm not sure if this is a real issue or just me, but UX wise this is not coincident and could confuse the user. What do you think?

lib/insert-link.js
josephfrazier: Oh, I hadn't considered that case. Yeah, it sounds like we should use the `tryLoad` logic in the live-resolver as well. I'll probably try to put together a pull request with your proposed syntax sometime this week or over the weekend.

josephfrazier:
That's a good point about UX inconsistency. I had initially threw together this PR as just a demonstration for https://github.com/OctoLinker/live-resolver/pull/12, but I definitely agree that the proposed functionality should be "all or nothing", applying to every link in the same way.

I'm starting to think the best way to do this is to copy _all_ resolution logic into live-resolver, so that we can always rely on having a single `href` that will work regardless of how many possible resolutions there are. Along those lines, maybe it would make sense to combine `live-resolver` into this repo, to make it easier to share logic between the client-side resolver and the server-side resolver? Alternatively, we could do away with the client-side resolution entirely

stefanbuck:
Sorry, busy days for me. Will respond to you beginning of next week.

Cheers Stefan

josephfrazier:
No worries, I've been busy as well. Enjoy the weekend!

stefanbuck:
Hi @josephfrazier,

while developing the latest version of the OctoLinker I had the same idea of moving everything to the `live-resolver`, but there a couple of issues with this approach.

If a user clicks on a file import statement like `import foo from './foo` the extension does a bunch of HEAD request to find the linked file. This is need, because especially for JavaScript the linked resource could be one of the following:

- `./foo.js`
- './foo/index.js`
- ./foo/index.jsx'
- './foo.json`

This is implemented in [resolver/javascript-file](https://github.com/OctoLinker/browser-extension/blob/5bbe5d829cb95b9fe127e0e81d19acf1827ffc6f/lib/resolver/javascript-file.js). 

Also if we would do this on the server, github would properly ban the server, because of too many requests from the same IP.

Another disadvantage of this approach is that it limits this features for public repositories. 

Finally this would generate a lot of traffic on my Heroku and if the server is down for some reasons the extension would be completely useless.


Your second idea of combining the live-resolver into this repo is also not perfect for the following reasons. The extensions would require more permissions in order to make requests to all the different registries. Right from the beginning I was aiming for less as possible permission for this extension to make it more attractive for the user. Further the idea of the current release is to support more and more languages and registries. Every time when a new language is added chrome would ask for new permissions before updating the extensions. Besides that, the user wouldn't benefit from caching already requested packages.

You see, it's not that easy to come up with a solution which address all the mention issues. Nonetheless , I'm grateful for your input!

josephfrazier:
Thanks for the explaining how the local file resolution works. I wasn't aware of that at the time, and it makes sense why that shouldn't be done on the server.

When I suggested combining the live-resolver into this repo, I didn't mean that all of the resolution for external packages should be done client-side (though that is an interesting idea). Instead, what I should have explained is that since the code is closely coupled across repos, it might make sense make sense to combine them into one repo. We could still deploy the extension and server separately, but if there were to be any shared logic, it would be easier to reuse code for it than if the extension code and the server code were in separate repos.

What if we were to add an "eager" mode to the extension? This would be disabled by default, but we could add an [options page](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Anatomy_of_a_WebExtension#Options_pages) for enabling it. If enabled, the extension would attempt to add an href to every link during the extension loading process, whether by resolving local files on the client-side (using the HEAD mechanism you described), or by linking to the server-side resolution (https://github.com/OctoLinker/live-resolver/pull/12) for external packages.

stefanbuck:
Which problem tries this "eager" mode to solve? However, this would put a lot of traffic to the service and even worse the initial load time would increase a lot. For example, a page with ten potential octolinker links and assuming each request takes ~300ms the total load time would be 3 seconds. 

Regarding the mono repo. This is kind of a mono repo already with the `/packages` directory so in general I'm open to extend it. However as you already mentioned, this only make sense if we share server code with the extension. I would like to turn off the server, but I'm not sure how beneficial this will be for us and our users. Why should we do that? Just for a proper link?

stefanbuck:
Closing this because of your comment in https://github.com/OctoLinker/live-resolver/pull/12#issuecomment-263438970

Issue #88, opened 2016-05-27
Doesn't work in conjunction with github-highlight-selected Chrome extension


dannyfritz:
It fails to redirect when `github-highlight-selected` is also installed https://github.com/Nuclides/github-highlight-selected

![atom_2016-05-27_08-44-13](https://cloud.githubusercontent.com/assets/315788/15611349/34495ee0-23e7-11e6-9686-242663349ac0.png)

Reason the bug is opened here instead of there is because the other extension works, but breaks this one. If you feel it is a problem on their end, tell me to open an issue there or open one with more specific details.

stefanbuck:
Thanks for filling this issue. The `github-highlight-selected` extension wraps every single word with a `<spam>` which cause this issue. Luckily the fix is quite simple on this end 😉 Updating the target source from `event.target` to `event.currentTarget` in the [click-handler](https://github.com/OctoLinker/browser-extension/blob/40cc8a5533e827361f42058731886b6c210f1d0a/lib/click-handler.js#L39) should fix this issue. Would you like to write a patch for this?

dannyfritz:
Yep, I can do a pull request sometime this weekend. :+1:

stefanbuck:
Will be fixed with the next release

PR #89, opened 2016-05-31
Use currentTarget for click handler event.

dannyfritz:
Happy Memorial Day!

This is related to item #88 

Using `currentTarget` in the click handler over `target` ensures that even if [other extensions](https://github.com/Nuclides/github-highlight-selected) change the DOM, we still get what we want.


stefanbuck:
Thanks 👍

Issue #86, opened 2016-05-27
Feedback on Contribution Process


josephfrazier:
(in response to https://github.com/OctoLinker/browser-extension/issues/72#issuecomment-219362198 and https://github.com/OctoLinker/browser-extension/pull/81#issuecomment-220519588)

Hey, I finally got around to collecting my thoughts on this. I hope they're helpful!

---

After searching around the codebase for a bit, it wasn't too hard to see how everything fit together. I ended up modeling the new `homebrew` plugin after the `gem` plugin, rather than the `bower` plugin you recommended, since the `gem` plugin was simpler. For future contributors, it might be nice to have a link in CONTRIBUTING.md or DEVELOPER.md (maybe combine the two?) to a "canonical" commit that shows an example of a plugin being added. I didn't think to look for one of these, and I'm sure it cost me a bit of time on understanding the architecture.

If I'm completely honest, though, the project feels a little over-engineered, from the perspective of someone who just started working on it. I'm guessing there's good reasons for why it's the way it is, but I'll go ahead and write out what I'm thinking anyway: It seems to me that adding a new pattern/resolver is a bit too complicated. As you can see, [the commit that added homebrew support](https://github.com/OctoLinker/browser-extension/commit/b6df755208eae0b98e55fdf533b647b5d7bd4f8e) ended up changing 4 files, whereas I feel like the project could have been architected to allow for something like this instead:

``` javascript
import relativeFile from './lib/resolver/relative-file.js'

const HOMEBREW = /(?:depends_on|conflicts_with) (['"][^'"\s]+['"])/g;

module.exports = {
  '*.rb': {
    // Using a RegExp as a property name is kinda sketchy, as it just gets
    // converted to its string representation. However, the module that imports
    // this could convert the string back into a RegExp
    HOMEBREW: function (path, matchResult) {
      // `path` is a string containing the path to the blob
      // `matchResult` is the return value of String.prototype.match
      const formulaName = matchResult[1]
      return relativeFile(path, formulaName + '.rb')
    },
    // Other patterns for .rb files go here
  },
  // Other filename patterns go here
};
```

I just wrote the above off the top of my head (and I'm running a bit low on sleep), so there might be some problems, but I hope it makes sense what I'm trying to get at. Let me know if I didn't explain enough, and I can fill in the blanks.

stefanbuck:
Thank you so much for your very useful feedback 👍 

[...]

PR #92, opened 2016-06-07
List example plugin commits in CONTRIBUTING.md

josephfrazier:
Addresses https://github.com/OctoLinker/browser-extension/issues/86#issuecomment-222381365:

>  - [ ] Example commit which show of a plugin being added.


stefanbuck:
👍

PR #94, opened 2016-06-07
Build chrome/app.js with source map

josephfrazier:
This lets developers debug the original code, rather than the compiled
output. Here's a before/after of the debugging experience:

Before:
before

After:
after

Note that this increases the size of chrome/app.js by roughly 1 MB. If that's a problem, We could use an [external source map](https://github.com/substack/node-browserify#external-source-maps) instead, and then figure out how not to include that in the published extension.

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -4,4 +4,4 @@ set -e
 NPM_BIN="$(npm bin)";
 cd "$(dirname "$0")/.."
 
-"$NPM_BIN/browserify" ./lib/app.js -t babelify --outfile ./chrome/app.js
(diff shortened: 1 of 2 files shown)


stefanbuck:
I'm sorry, but I want to replace browserify with webpack soon. I already started https://github.com/OctoLinker/browser-extension/commits/webpack, but this PR based on https://github.com/OctoLinker/browser-extension/commits/firefox which is currently blocked, because I'm waiting for feedback from Mozilla.

stefanbuck:
i just realized that karma is still using `browserify`. Will work an that later in the day.

stefanbuck:
Closing this in favor of https://github.com/OctoLinker/browser-extension/pull/97.

Sorry @josephfrazier 😉

Issue #93, opened 2016-06-07
`require.resolve('os-browserify/browser.js')` links incorrectly


josephfrazier:
To reproduce the bug, visit https://github.com/substack/node-browserify/blob/918d62e804e0e735708e24fe53a78bca0400b7e5/lib/builtins.js#L17 and click on 'os-browserify/browser.js', then get taken to https://github.com/springload/Browser.js.

Instead, the link should go to https://github.com/CoderPuppy/os-browserify (or even https://github.com/CoderPuppy/os-browserify/blob/master/browser.js, but that's a little more complicated...)

I'll take a look at fixing this, but I wanted to make sure there was an issue for it, in case I'm not able to get it working.

josephfrazier:
>  (or even https://github.com/CoderPuppy/os-browserify/blob/master/browser.js, but that's a little more complicated...)

I was just thinking about how this might work, when I realized that I had tried it [before](https://github.com/OctoLinker/live-resolver/pull/3) 😝

josephfrazier:
I'm closing this, since it was effectively resolved by https://github.com/OctoLinker/browser-extension/commit/e904209a9eff158399317be1e237fb13876ccb18. There's still the possibility of "deep linking" paths within packages as described in my previous comment, but that's a separate (and much more complicated) problem.

PR #95, opened 2016-06-08
npm: resolve 'foo/bar.js' as if it were 'foo'

josephfrazier:
Addresses https://github.com/OctoLinker/browser-extension/issues/93


lib/resolver/live-resolver-query.js
```diff
@@ -1,4 +1,9 @@
 export default function ({ type, target }) {
+  if (type == 'npm') {
+    // If the target looks like 'foo/bar.js', pretend it is 'foo' instead. See
+    // https://github.com/OctoLinker/browser-extension/issues/93
+    target = target.split('/')[0];
+  }
```
stefanbuck: Mhhhh I'm not sure if this is the right place to do this. Maybe a `shorthand-resolver` ... let me think about it  ...

stefanbuck:
Regarding https://github.com/OctoLinker/live-resolver/pull/3#issuecomment-141370198, I would prefer to start simple with the issue described in #93. You can ignore the scoped package and duojs issue for now.

lib/resolver/live-resolver-query.js
stefanbuck: something like that, in a super simple way.

```js
export default function ({ type, target }) {
  if (type === 'npm') {
    let repoTarget = target.split('/')[0];
    return [
      liveResolverQuery({ type, repoTarget }),
    ];
  }

  return [
    liveResolverQuery({ type, target }),
  ];
}
```

josephfrazier: Hmm, yeah, I also wasn't completely sure where it should go. My logic at the time was to put it as close to the server as possible, in case we wanted to move the logic into live-resolver in order to account for the edge cases mentioned in https://github.com/OctoLinker/live-resolver/pull/3#issuecomment-141370198

However, I don't yet understand why it should be a separate resolver. To me, that seems like an unnecessary layer of complexity, but perhaps there are some cases I haven't considered that would be broken by changing the `liveResolverQuery` directly.

At any rate, I've updated the PR with some changes inspired by your snippet above, as well as a test. Let me know what you think!

stefanbuck: I think, the `liveResolverQuery` shouldn't know anything about npm or something else. He should care about building the live resolver request url based on the passed args. Imagine we want to have the same behaviour for another languages lets say PHP. In that case we would have to pollute the `liveResolverQuery` with logic for handling PHP. That's my concern, why it shouldn't be in here.

stefanbuck:
What happened with the comments in here? Can you see them @josephfrazier ?

lib/resolver/live-resolver-query.js
josephfrazier: Oh, that makes sense. I've moved the logic into `javascript-universal.js`. Does that seem like an appropriate place for it?

josephfrazier:
Hey, any thoughts on https://github.com/OctoLinker/browser-extension/pull/95#discussion_r67629777 above? It looks like the github UI is hiding it, so we should probably just continue the conversation here.

stefanbuck:
Sorry, for the delay. I see your messages when I wake up and reply mostly in the evening 😉 Anyway, this looks much better. We should release a new version soon. Good work 👍

stefanbuck:
Please rebase and push the button

josephfrazier:
No problem about the delay. This will need to be adapted to correctly handle scoped modules, but I went ahead and merged because I think that `require('non-scoped-module/file.js')` is probably more common than `require('@scoped/module')` (which didn't work until https://github.com/OctoLinker/live-resolver/pull/13 was released, anyway). I'll probably start working on a fix for this soon.

PR #97, opened 2016-06-08
Replace browserify with webpack

stefanbuck:
I'm a big fan of both worlds, but I think webpack gives us more flexibility and freedom. Beside adding webpack this PR replaces most of the bash scripts with npm script, because this is more  common.

@josephfrazier can you review it, please.


josephfrazier:
I'm not especially familiar with webpack, but everything still works, so this looks good to me! I'm especially a fan of moving the bash scripts into package.json. Along those lines, I'd suggest going one step further and moving `scripts/build` as well. Here's a commit that does just that: https://github.com/josephfrazier/browser-extension/commit/0e11b3e66b34263a0a0a1ca5f041ed45464b0f42

stefanbuck:
I dropped the build script in the firefox PR https://github.com/OctoLinker/browser-extension/pull/98/commits/281eb079c442c2313ea32246ee30efeed3c9a9db. But you're right should be go in with this PR.

josephfrazier:
Oh, I hadn't looked at the firefox PR yet, but thanks for pulling in my changes. One of the Travis build jobs had initially failed due to network connectivity issues, but it worked after I restarted it. If I had to find nits to pick, I'd ask why [`__dirname` stopped working](https://github.com/OctoLinker/browser-extension/pull/97/files#diff-3451d53a8bff0930e799ae5a156cbee0R13), but that definitely shouldn't block this being merged in 😄

josephfrazier:
I was just about to merge this, per our twitter messages, but I noticed a problem. The postinstall script fails if the pre-commit hook is already installed. This is fixed by https://github.com/josephfrazier/browser-extension/commit/98b1e8e27247370560b72f43560bfade257ebc66

PR #99, opened 2016-06-10
postinstall: overwrite pre-commit hook if needed

josephfrazier:
This lets you `npm install` multiple times without issue.

(from https://github.com/OctoLinker/browser-extension/pull/97#issuecomment-225069078)


stefanbuck:
:+1:

PR #100, opened 2016-06-15
Add Firefox support

stefanbuck:
This PR adds the Firefox version of the OctoLinker extension. To build the Firefox version run `npm run build-firefox`. To build a release version run `npm run dist-firefox` or `npm run dist-all` to build Chrome + Firefox.

In order to make the Firefox version working some changes in the request layer were needed. Firefox behaves a bit different in terms of [CSP](https://developer.mozilla.org/en-US/docs/Web/Security/CSP) than Chrome. By that, a request to the [OctoLinker server](https://githublinker.herokuapp.com/) wasn't possible from the [content script](http://stackoverflow.com/a/12972014). Luckily the reuqest can still be made with a background script. Therefore I dropped the `try-load` package and move the logic from there to `background.js`. To keep it consistent and easy, all requests (even in Chrome) are using the background script to load data.


josephfrazier:
I haven't set up Firefox for testing this yet (maybe CONTRIBUTING.md should link to https://developer.mozilla.org/en-US/Add-ons/Setting_up_extension_development_environment?), but this looks good to me otherwise!

A couple of future improvements come to mind:

* Can we generate `firefox/manifest.json` from `chrome/manifest.json`, or vice versa? That way, we wouldn't have to manually keep them in sync.
* Now that there are dist builds, the normal builds should be able to include sourcemaps (see https://github.com/OctoLinker/browser-extension/pull/94). The dist build could remove the sourcemap files, so that they don't get packed into the released extensions.

josephfrazier:
I got my Firefox development environment set up (see https://github.com/josephfrazier/browser-extension/commit/e76e4f1f6db3163a64ff4bfb61e42d357d22ef1d) and found/fixed a bug. See https://github.com/josephfrazier/browser-extension/commit/a52c771c7da4e77453694054072395095267e765


> Ensure chrome.runtime.onMessage.hasListeners exists before calling it

> Without this, I was seeing the extension hang immediately after clicking
on a link, and the following error was printed out:

>     Firefox stderr: JavaScript error: resource://gre/modules/ExtensionContent.jsm -> moz-extension://3f01c148-190b-214c-8200-657ba756f784/app.js, line 18560: TypeError: chrome.runtime.onMessage.hasListeners is not a function

stefanbuck:
First, thanks for the review @josephfrazier.

I didn't even know that `web-ext` exists, good spot. The bug that you found wasn't a problem for me during the development, so I guess this is somehow related to `web-ext`. I have to investigate that.

Regarding the manifest files. I agree that maintaining two files isn't ideal. I was quickly search for a webpack plugin which allows to modify the content of a file while copying it from a to b, without luck. Maybe it's time to add a task runner like `grunt` or `gulp` whatever fits better. I would like to ignore this for now. We can it improve and discuss it in a separate PR.

Finally, indeed sourcemaps are great and super helpful, let's implement them.

stefanbuck:
@josephfrazier found the problem why the extension was hanging after a click.  `chrome.runtime.onMessage.hasListeners` is not supported in the WebExtension API. Removed the function and implemented a flag to determine if the eventHandler is already added or not.

stefanbuck:
@josephfrazier please, merge this pull request if you think it's ready.

firefox/manifest.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "OctoLinker",
+  "version": "4.1.1",
```
josephfrazier: Should we bump the version in chrome/manifest.json as well? I'm assuming they're meant to be synchronized.

josephfrazier:
Thanks for the reminder. Other than the above question about version numbers, this looks good to me. 

I hadn't heard of web-ext before either, but it seems to help a lot with the Firefox extension development process, which I wasn't previously familiar with. Thanks for adding in my `open-firefox` commit, as well as the sourcemaps! Glad to see a proper fix for the `hasEventListeners` thing as well (my "fix" for it was just the easiest thing that didn't throw an error, but I wasn't really paying attention to how it should have been done :P)

I'll go take a look at #101 for the manifest file generation.

firefox/manifest.json
stefanbuck: I would say no, because there is no chrome version 4.1.1 available. In future they will be always in sync, I guess.

Regarding the `version` field in the `package.json`. I think we can get rid of it completely since this isn't a npm module.

josephfrazier: Oh, ok. Maybe we're thinking about these version numbers differently. I'll go ahead and merge this PR, then continue the conversation in #101

PR #101, opened 2016-06-18
Master manifest file

stefanbuck:
**Work in progress and branched off from https://github.com/OctoLinker/browser-extension/pull/100**

Based on the comment https://github.com/OctoLinker/browser-extension/pull/100#issuecomment-226648352 this PR adds gulp which allows us to extend the manifest file with additional information.

New commits on top of #100 are
- 5102bca22b9f6b6fe722ed868b59efda7bb1bc62
- 715abdf04ac8c2c27962c31fc1e9603199fddac3
- 55b20755051ff0a31b0ac3edcd9fb36060fa12f9
- 3dfa20f660381f37d319627a81c926a7c1d78b98


josephfrazier:
Wow, I wasn't expecting so much effort to go into generating the manifest files, but it looks like there's some other changes here too. I'm wondering if we really need gulp and all the reorganization, though. For example, we could generate `chrome/manifest.json` from `firefox/manifest.json` with this npm script (using the [`json` package](https://www.npmjs.com/package/json)):

```javascript
"build-chrome-manifest": "json -e 'delete this.applications' < firefox/manifest.json > chrome/manifest.json",
```

Likewise, the `pack-repo` task could be simplified to this npm script:

```javascript
"pack-repo": "mkdir -p out && zip --recurse-paths out/octolinker-repository-$npm_package_version.zip * --exclude dist/\\* out/\\* node_modules/\\*",
```

What do you think?

stefanbuck:
Mhhh to be honest, I'm a bit prejudiced, as I worked a couple of hours to get it working 😄  I see you point and in general I totally agree, but maybe there is some good reason to use gulp – let me think ...

stefanbuck:
The firefox mainfest.json needs an additional `permissions` property `https://github.com` and chrome don't. Both need the permission `https://githublinker.herokuapp.com/`. How can we achieve that with https://www.npmjs.com/package/json?

josephfrazier:
Oh, I definitely understand not wanting to let your work go to waste 😄 I certainly have no problem with using gulp, but it just struck me as overkill.

Nice catch on the `permissions` difference between firefox/chrome manifest.json. I wasn't sure if that was intentional or not. Anyway, we could remove it from the chrome manifest.json by changing the npm script to this:

```javascript
"build-chrome-manifest": "json -e 'delete this.applications; this.permissions.shift()' < firefox/manifest.json > chrome/manifest.json",
```

To tie in what I was saying about the package.json version in https://github.com/OctoLinker/browser-extension/pull/103#issuecomment-227255630, here's a script that copies the `version` property from package.json to firefox/manifest.json:

```javascript
"update-firefox-manifest": "json -I -f firefox/manifest.json -e \"this.version = '`json version < package.json`'\"",
```

stefanbuck:
I'm gonna close this in favor of #107

PR #103, opened 2016-06-20
Remove unnecessary fields from package.json

stefanbuck:
name, version and main are only needed for publish a module to the npm registry. Since this will never be published to npm we can just remove them to avoid any confusion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,7 @@
 {
-  "name": "OctoLinker",
-  "version": "4.1.0",
   "engines": {
     "node": ">=4.0.0"
   },
-  "main": "lib/app.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


josephfrazier:
Along the lines of what I've suggested in https://github.com/OctoLinker/browser-extension/pull/101#issuecomment-227030823, what if we instead treat the package.json version as the single source of truth, and copy it into the chrome/firefox manifest.json when doing dist builds? If we leave it in package.json, it would also be available to npm scripts as the `$npm_package_version` variable, as seen in the `pack-repo` script proposed in https://github.com/OctoLinker/browser-extension/pull/101#issuecomment-227030823

stefanbuck:
Valid points, will close this because of the ongoing discussion in #101

Issue #29, opened 2015-02-16
Safari support


tjmehta:
I love this plugin. I would love to see safari support. Thanks!

tjmehta:
Octotree, another browser plugin I love has recently started supported all browsers. It looks like a good examples of how to support multiple plugin formats ( Firefox too #24). Check out the source here: https://github.com/buunguyen/octotree

stefanbuck:
The problem is not to build these extensions, but keep them up to date with the registries. Accutlay once per week I run a script which generates the cache list for the npm, bower and comperser registry. 

I'm planning to move the core from a cache list to a live resolver solution, but I need time to develop it. Soon as this is done I can start thinking about a safari and firefox version :wink:

tjmehta:
Ah, I was wondering what you were doing the linking. Is it not possible to do everything clientside (maybe parse npmjs or bower websites for the github links - or maybe reverse engineer their apis)?

stefanbuck:
Yeah, that's the plan. I will start working on it soon :wink:

JaKXz:
Any updates on the safari extension? The developer preview for Safari 10 is actually pretty good, IMO :)

stefanbuck:
Nope, unfortanly not. Do you know if Safari 10 supports the [WebExtension API](https://wiki.mozilla.org/WebExtensions)?

josephfrazier:
I looked into this for a little bit, and found that while Safari doesn't support the WebExtension API, it might not be very difficult to shim it out, using something like https://github.com/kritollm/chrome-extension-api-for-safari-and-firefox

Another potential option could be to build a userscript version of the extension, so Safari users could install it using [Tampermonkey](https://tampermonkey.net/?browser=safari), etc

nikitavoloboev:
I too would love for Safari to be supported

sindresorhus:
Safari will soon support the WebExtensions API: 
- https://hacks.mozilla.org/2020/06/welcoming-safari-to-the-webextensions-community/
- https://developer.apple.com/documentation/safariservices/safari_web_extensions/converting_a_web_extension_for_safari

stefanbuck:
Thanks for the update. Do you know if a developer account is required to submit web extensions? The developer program is super pricey to distribute a single free item.

nesl247:
Yes, it is still required. Though if you already have one for any purpose, you could use that.

I'd be willing to chip in to help pay for it.

fregante:
[...]

PR #107, opened 2016-06-21
Use npm script to build and pack extension

stefanbuck:
Based on the discussion with @josephfrazier on #101 this PR use npm script instead of gulp for building  and packing the extension. Except `npm run *-pack` it works quite well. I'm just worried about the complexity in the `package.json` file. I'm undecided which version I prefer more. 

@josephfrazier regarding the issue with `npm run *-pack`. When I run the script in the console it works, but the same script execute via npm script returns 

```
zip error: Nothing to do! (out/chrome-octolinker-4.1.0.zip)
```

Any ideas?


josephfrazier:
Hmm, when I run `npm run chrome-pack`, it fails with this message:

```
npm ERR! Tell the author that this fails on your system:
npm ERR!     npm run chrome-dist && mkdir -p out && zip --recurse-paths out/chrome-octolinker-$npm_package_version.zip dist/ --exclude dist/ \* out/\* node_modules/\*
```

But when I run the suggested command above, it also fails with:

```
zip error: Nothing to do! (out/chrome-octolinker-.zip)
```

(Note that `$npm_package_version` is empty, because we're not running in the context of an npm script.) 

It looks like the `chrome-pack` command was modeled after `pack-repo`, but since it's only including the files in `dist/`, there's no need to `--exclude` anything. In particular, `dist/` is being excluded, which explains why `zip` says there's "Nothing to do!". Maybe `chrome-pack` should be this instead?

```
npm run chrome-dist && mkdir -p out && zip --recurse-paths out/chrome-octolinker-$npm_package_version.zip dist/
```

stefanbuck:
@josephfrazier please have a look again. TA

josephfrazier:
Ah, that looks much better! Nice find on the `zip -j` flag to avoid having `dist/` go into the zipped paths; I wasn't aware of that.

I do have one nitpick, though: My development process involves loading the unpacked Chrome extension directly from `./chrome`, but with these changes, `npm run chrome-build` doesn't actually create `./chrome/manifest.json`, so the extension fails to load. I started thinking about how to fix this, and I realized that we could do away with the `dist/` directory entirely. Does that sound ok to you? Here's a commit for it: https://github.com/josephfrazier/browser-extension/commit/b5d5300bb0ed99ca40d18674a7f14ed6a1c566d7

stefanbuck:
My idea was to get rid of the `chrome/` and `firefox/` directory and use `dist/` instead for development. Maybe it's just a naming issue. What about `build`? The only disadvantages I can think of is that pair development in Firefox and Chrome is not possible anymore, which shouldn't be a big issue. Btw `npm run chrome-build ` produces a `manifest.json` file, under `dist/` and not as you may expected under `chrome/`.

josephfrazier:
Oh, I see now. I was under the impression that the chrome and firefox directories were still used somewhere in the build process, but I now see that I was mistaken. I think it's cleaner to have a single output directory anyway, especially if we were to add [support for other browsers like Safari](https://github.com/OctoLinker/browser-extension/issues/29). Merging :)

stefanbuck:
What about the naming? Should we call it `build` instead of `dist`?

josephfrazier:
I don't really have a preference there, other than a very slight preference to leave it as `dist`, so I don't have to change the path that Chrome loads the unpacked extension from :P

PR #108, opened 2016-06-27
Plugins located in lib/plugins will be auto-loaded

stefanbuck:
Based on the feedback from @josephfrazier in https://github.com/OctoLinker/browser-extension/issues/86, this PR adds an auto-loader which loads all plugins located in `lib/plugins`. By that, a plugin developer doesn't have to think about import the plugin manually.


josephfrazier:
Nifty! I had heard that webpack can do dynamic requires like this, but I don't think I'd seen it used until now.

Issue #106, opened 2016-06-21
Does not linkify extension-less files


josephfrazier:
For example, lines 2 and 3 of this blob should have links, but they don't, since the blob name has no extension: https://github.com/localtunnel/localtunnel/blob/9a1d48764a167ae9fb05f42efdd7de0005e9edfc/bin/client

Perhaps we should add a way to detect filetype based on the blob's [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix))? In the above example, `#!/usr/bin/env node` indicates that the file should be interpreted as javascript.

stefanbuck:
Right, there is some space for improvements. I like the idea of guessing the filetype by looking at the blob type. 

Similar to https://github.com/OctoLinker/browser-extension/issues/56

PR #109, opened 2016-06-27
WIP: Add shebang-based filetype detection

josephfrazier:
This is a relatively quick and dirty fix for https://github.com/OctoLinker/browser-extension/issues/106

There's still some work to be done (adding tests, adding more flexible shebang patterns, etc), but I wanted to get some feedback on the approach before putting more work into it. What do you think?


stefanbuck:
Mhhh, to be honest, I don't think this is the right way. What happens if there is no Shebang present or it is slightly different? Anyway, GitHub does the type selection for us. Most of the blobs have a css class which is a reference to the file type like `type-javascript` or `type-shell`. One of the first `blob-reader` version was using this css class to detect the file type. At this time, I was thinking that dealing with file extensions is easier than a blob type. Maybe we should provide both for every blob.

```html
<div class="blob-wrapper data type-javascript">
```

josephfrazier:
Oh, I wasn't aware of github's type detection. Using that sounds much easier than parsing a shebang that may not exist in the first place. I've refactored my existing work to do that instead. What do you think?

stefanbuck:
Should we really keep it that plain to the developer? Would like to merge the class name somehow into `getPattern()`. Ideally a [patter preset](https://github.com/josephfrazier/browser-extension/blob/926b25215315366757c593b3cf68eae62ae9055a/lib/pattern-preset.js) would know which css class is associated with a language. Do you know what I mean?

josephfrazier:
Oh, I think I see what you mean. [`getPattern()`](https://github.com/OctoLinker/browser-extension/blob/d5f1936373571e422743bbb7d30893da84b4e6c3/lib/plugins/javascript/index.js#L7) could return an object like `{pathSubstrings: ['.js', '.jsx', '.es6'], githubClass: 'type-javascript'}`, and [`buildPluginCache`](https://github.com/OctoLinker/browser-extension/blob/d5f1936373571e422743bbb7d30893da84b4e6c3/lib/plugin-manager.js#L4) could be adapted to build a separate cache for the `githubClass` values. Does that sound better?

stefanbuck:
Sounds good:+1:

josephfrazier:
Great, I just pushed some changes that move the github classes into getPattern() (and the pattern presets where possible). I thought about making a separate Map for associating github classes with plugins, but I decided to just reuse the same one for now. What do you think?

lib/plugins/bower-manifest/index.js
```diff
@@ -27,7 +27,9 @@ function linkFile(blob, key, value) {
 export default class BowerManifest {
 
   getPattern() {
-    return ['bower.json'];
+    return {
+      pathSubstrings: ['bower.json'],
+    };
```
stefanbuck: Mhhh I think there is no case where a plugin needs to return a GitHub-Class right? Therefore I would like to keep the plugin interface like it was before and do the GitHub-Class lookup behind the scenes in the `plugin-manager`.

lib/pattern-preset.js
```diff
@@ -1,10 +1,13 @@
...
-    '.es6',
-  ],
+  'JavaScript': {
+    pathSubstrings: [
+      '.js',
+      '.jsx',
+      '.es6',
+    ],
+    githubClasses: ['type-javascript'],
+  },
```
stefanbuck: According to my comment below, keeping the classname here is fine because it's the right place for it.

lib/plugin-manager.js
```diff
@@ -1,26 +1,38 @@
...
 
-    plugin.getPattern().forEach((pattern) => {
-      const pluginsForPattern = lookup.get(pattern) || [];
-
-      if (pluginsForPattern.includes(plugin)) {
-        return;
-      }
-
-      pluginsForPattern.push(plugin);
+    plugin.getPattern().pathSubstrings.forEach((pattern) => {
```
stefanbuck: This is the right place where we can distinguish between the plugin interface (an array) and the pattern-preset object interface.

stefanbuck:
Added some comments which are maybe related to your Map suggestion (I'm not sure). Can you quickly explain your idea, because I don't get it. Thanks

stefanbuck:
Do you need help with this?

josephfrazier:
Hey, I hadn't made time for this recently, but I think I can get back to it now. I see your 3 comments above, but it sounds like you're saying you want `plugin.getPattern()` to go back to returning only an array of strings, instead of including the `githubClasses`, which seems to contradict your [earlier comment](https://github.com/OctoLinker/browser-extension/pull/109#issuecomment-229796293). I'm not quite sure what you're suggesting, could you can push a commit somewhere with the architecture you have in mind?

stefanbuck:
Basically everything looks good. What I meant in the comment before is, that the return value of  `getPattern` shouldn't change. A plugin will always return an array of `pathSubstrings` so the object is just an unnecessary layer there. To stop any further confusion on that from my side I've created a example commit as requested from you 😉  https://github.com/stefanbuck/browser-extension/commit/0396e7b5ef48675e38ac5312e929adb7b2dfc058 Sorry for the confusion.

josephfrazier:
Ohhh, I see now. You want the plugin manager to be backward-compatible with plugins whose `getPattern()` return value is the same as it was before. I've pushed some changes that make this branch match up with your example. Could you take another look?

stefanbuck:
Looks good. Just one tiny thing. Can you add the css classes for the other languages as well, please.

lib/plugins/docker/index.js
```diff
@@ -4,7 +4,10 @@ import insertLink from '../../insert-link';
 export default class Docker {
 
   getPattern() {
-    return ['Dockerfile'];
+    return {
+      pathSubstrings: ['Dockerfile'],
+      githubClasses: ['type-dockerfile'],
+    };
```
stefanbuck: What was the reason for not adding this to our `pattern-preset`?

lib/plugins/vim/index.js
```diff
@@ -4,15 +4,18 @@ import insertLink from '../../insert-link';
...
+      pathSubstrings: [
+        '.vimrc',
+        '.gvimrc',
+        '.vim',
+        // Sometimes there's no leading dot in .vimrc and .gvimrc, for example:
+        // https://github.com/gmarik/vimfiles/blob/1f4f26d42f54443f1158e0009746a56b9a28b053/vimrc#L136
+        'vimrc',
+        'gvimrc',
+      ],
+      githubClasses: ['type-viml'],
```
stefanbuck: Same as above

lib/pattern-preset.js
```diff
@@ -1,18 +1,27 @@
...
 
-  'TypeScript': [
-    '.ts',
-  ],
+  'TypeScript': {
+    pathSubstrings: ['.ts'],
+    githubClasses: [
+      'type-typescript',
+    ],
+  },
```
stefanbuck: Can not comment on the live below, but can you add the `type-ruby` class there. TA

josephfrazier: Oops, not sure how I missed that one. I'll add it in, along with `type-coffeescript` above.

lib/plugins/vim/index.js
josephfrazier: I'll go ahead and move them over :)

stefanbuck:
LGTM 👍

josephfrazier:
Woot! Glad to get this one landed, thanks for all the help.

Issue #110, opened 2016-06-27
unable to resolve link


zeke:
![screen shot 2016-06-27 at 4 48 20 pm](https://cloud.githubusercontent.com/assets/2289/16399410/0a87659e-3c87-11e6-9287-a75b5f4b72b3.png)

zeke:
https://github.com/bfred-it/npm-hub/blob/507d1c2980810a28b7bff82af2e7f384245446f7/package.json

zeke:
https://www.npmjs.com/package/chrome-webstore-upload-cli

josephfrazier:
This is because https://registry.npmjs.org/chrome-webstore-upload-cli doesn't have any of the fields defined in [the live-resolver configuration](https://github.com/OctoLinker/live-resolver/blob/648a6ee493d46f7563ae0c06d4b3709911d15c79/config.json#L28-L41). Namely, the [`repository`](https://docs.npmjs.com/files/package.json#repository) field is missing from [the chrome-webstore-upload-cli package.json](https://github.com/DrewML/chrome-webstore-upload-cli/blob/e196dd5014463016ebd7fd5e2e50fb785912cf48/package.json).

josephfrazier:
Oh, I see that while the aforementioned live-resolver configuration has [a fallback url defined](https://github.com/OctoLinker/live-resolver/blob/648a6ee493d46f7563ae0c06d4b3709911d15c79/config.json#L27), it's not actually being used. I'll put together a fix for that.

zeke:
Sweet, thanks @josephfrazier.

I also went ahead and added the repo to that project's package.json: https://github.com/DrewML/chrome-webstore-upload-cli/pull/6

josephfrazier:
Sure thing, thanks for the report. Also, for introducing me to `chrome-webstore-upload-cli`. That looks useful!

PR #112, opened 2016-06-28
Upload builds to GitHub releage page

stefanbuck:
Every time when a new git tag is pushed to master travis will automatically build and upload the browser builds to the corresponding git tag on the release page

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,14 @@ node_js:
   - 4.0
   - 4
   - 6
+after_success:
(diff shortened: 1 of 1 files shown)


josephfrazier:
👍 We might even be able to go a step further and use something like [`chrome-webstore-upload-cli`](https://www.npmjs.com/package/chrome-webstore-upload-cli) (from https://github.com/OctoLinker/browser-extension/issues/110) to automatically publish (not sure about Firefox, though).

PR #114, opened 2016-06-30
Insert <a> (no href) instead of <span>

josephfrazier:
This allows the links to be "clicked" with [Vimium].

[Vimium]: https://vimium.github.io/

(cherry-picked from commit 0a69feedd80a39197b9cf7aa0b9bdf88eee56892)


stefanbuck:
The link color is slightly different. Please add the following css to ensure that the syntax highlighting is still the same. 

```css
a.octo-linker-link {
    color: inherit;
}
```

josephfrazier:
Good eye, thanks!

PR #115, opened 2016-06-30
Correctly resolve targets like @angular/core/bar.js

josephfrazier:
See https://github.com/OctoLinker/browser-extension/pull/95#issuecomment-229799358

### DIFF
--- a/lib/resolver/javascript-universal.js
+++ b/lib/resolver/javascript-universal.js
@@ -3,6 +3,13 @@ import javascriptFile from './javascript-file.js';
 import liveResolverQuery from './live-resolver-query.js';
 import liveResolverPing from './live-resolver-ping.js';
 
+function getTopModuleName(target) {
+  const isScoped = target.startsWith('@');
+  const numComponents = isScoped ? 2 : 1;
+  const topModuleName = target.split('/').slice(0, numComponents).join('/');
+  return topModuleName;
+}
+
 export default function ({ path, target }) {
   const isPath = !!target.match(/^\.\.?[\\|\/]?/);
   const isBuildIn = builtins.indexOf(target) > -1;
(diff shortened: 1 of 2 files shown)


josephfrazier:
Hmm, Travis failed on the build against my repo: https://travis-ci.org/josephfrazier/browser-extension/jobs/141482374#L1086

This particular failure has happened before, in this repo: https://travis-ci.org/OctoLinker/browser-extension/jobs/140987735#L1098

        AssertionError: ' // Most popular rabbit names' == '// Most popular rabbit names'
            at fail (webpack:///~/assert/assert.js?9281:200:0 <- test/main-packages.js:10454)
            at equal (webpack:///~/assert/assert.js?9281:224:0 <- test/main-packages.js:10478)
            at webpack:///packages/blob-reader/test.js:157:8 <- test/main-packages.js:11568

I know I've seen it another time, and I restarted the build to fix it, but I'm still wondering what's going on here...

lib/resolver/javascript-universal.js
```diff
@@ -3,6 +3,13 @@ import javascriptFile from './javascript-file.js';
 import liveResolverQuery from './live-resolver-query.js';
 import liveResolverPing from './live-resolver-ping.js';
 
+function getTopModuleName(target) {
+  const isScoped = target.startsWith('@');
```
stefanbuck: Is it really that simple to differentiate between scoped packages and normal packages? Are there no other edge-cases? At least normal npm packages can't start with an @-sign, but what about others package manger? Maybe I'm thinking it over ...

stefanbuck:
Beside one tiny comment, looks good to me.

Regarding the travis issue. That's super strange. I'm wonder, why it's only failing for the first time, only on node 6 and only on travis? I tried to reproduce it local with node 6, but there it worked. Investigate further.

lib/resolver/javascript-universal.js
josephfrazier: Hmm, I'm not too familiar with other package managers. In my opinion, this is at least an incremental improvement upon the current state of things. If we later discover that other edge cases are not handled correctly, we can fix those then. Does that sound ok to you?

josephfrazier:
Thanks, I responded to the tiny comment above.

My best guess on the travis issue (without having looked at it recently) is that it's a timing bug in the test's interaction with phantomjs. Not sure if it's really worth digging into at the moment, since restarting the test seems to be enough to fix it.

stefanbuck:
Please rebase and merge. Thanks for the improvement!

PR #119, opened 2016-07-06
Add auto publisher 

stefanbuck:
With this PR a git tag push will automatically build and upload a new version of the extension to the Chrome Web Store. 

### Todo
- [x] Add following travis environment variables `CLIENT_ID `, `CLIENT_SECRET `, `REFRESH_TOKEN` and `EXTENSION_ID`


josephfrazier:
This mostly looks good to me! One nitpick I have is that it seems that people other than yourself (namely, me) can push tags to this repo, and thereby cause releases to happen. I'm not sure if Github lets you restrict tag pushes, but it might be worth considering, especially if you don't trust me or future contributors not to hijack the extension :P

stefanbuck:
I think you should be able to push tag into master. Let's try that out for the next release.

PR #123, opened 2016-07-14
Don't test Node v4.0 on Travis

josephfrazier:
Travis already tests Node v4 at the latest minor/patch version.


stefanbuck:
Good spot 👍

PR #124, opened 2016-07-14
Support Homebrew-Cask dependencies/conflicts

josephfrazier:
This is only for `cask:` dependencies/conflicts. Here's an example:

https://github.com/caskroom/homebrew-cask/blob/d58e2c0724fe782a4ba18383e3dc387c59d63567/Casks/lazarus.rb#L11


stefanbuck:
Looks good to me, besides missing tests for the new behaviour 😉  Thank you

Issue #67, opened 2016-03-20
Support TypeScript reference tag


stefanbuck:
## Example
`<reference path="Validation.ts" />`

## Spec
https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#1111-source-files-dependencies

jamestalmage:
Also, require statements in `.ts` files should just work.

https://github.com/unional/ava/blob/master/index.d.ts#L1

stefanbuck:
Added TypeScript support in the latest version `v4.2.0`

PR #130, opened 2016-07-26
Add Typescript support

stefanbuck:
screen shot 2016-07-30 at 01 15 55

This PR adds support for TypeScript `import` and `require` statements. Also it allows to click on the path specified in a ts reference tag see #67.

In order to make this happen I had to tweak the `insert-link`, because the html structure for ts files is a bit different than for javascript. 

@josephfrazier  can you also do a quick sanity check with some TS project, please. That would be great.

### DIFF
--- a/lib/insert-link.js
+++ b/lib/insert-link.js
@@ -78,6 +78,18 @@ function getIndexes(portion, entireMatch, matchValue) {
   };
 }
 
+function wrapClosestElement(node, dataAttrObject, matchValue) {
+  let currentNode = node;
+
+  while (!currentNode.textContent.includes(matchValue)) {
+    currentNode = currentNode.parentNode;
+  }
+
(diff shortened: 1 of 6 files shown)


josephfrazier:
Nice work! I had a local branch with a small amount of work towards this, but I hadn't yet gotten it working. I tested against https://github.com/Microsoft/TypeScript/blob/730f18955dc17068be33691f0fb0e0285ebbf9f5/scripts/bisect-test.ts and the `require` calls worked well, but the reference tag did not.

stefanbuck:
Thanks @josephfrazier, good catch. It wasn't that easy as expected in the beginning to extend the linking mechanism.

josephfrazier:
Good stuff! Yeah, it does look like it ended up being rather tricky...

PR #131, opened 2016-07-29
Remove HTML files from GH stats

stefanbuck:
The stats report for the repository aren't correct, because of the massiv fixtures files for the blob reader. To fix this, I added an entry to `.gitattributes` to remove thoese files from the stats. For more details go to https://github.com/github/linguist#overrides

screen shot 2016-07-29 at 23 53 22

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 * text=auto
+packages/blob-reader/fixtures/* linguist-documentation


stefanbuck:
FYI @josephfrazier

stefanbuck:
Looks much better now 
screen shot 2016-07-30 at 00 02 34

josephfrazier:
Huh, I didn't know you could do that. Neat!

PR #134, opened 2016-07-30
homebrew: use homebrew-core for fallback link

josephfrazier:
This makes it so that homebrew formulae in other taps are properly
linked to their dependencies in the default tap. For example:

https://github.com/Homebrew/homebrew-science/blob/1acf4f470fc8c87f6bcbf19b721ebcd09c7fb025/octave.rb#L44

shows a dependency on autoconf, but the relativeFile resolver points to:

https://github.com/Homebrew/homebrew-science/blob/1acf4f470fc8c87f6bcbf19b721ebcd09c7fb025/autoconf.rb

which does not exist. Instead, use a new resolver to fall back to:

https://github.com/Homebrew/homebrew-core/blob/master/Formula/autoconf.rb

### DIFF
--- a/lib/plugins/homebrew-manifest/index.js
+++ b/lib/plugins/homebrew-manifest/index.js
@@ -20,7 +20,7 @@ export default class HomebrewManifest {
 
   parseBlob(blob) {
     insertLink(blob.el, HOMEBREW, {
-      resolver: 'relativeFile',
+      resolver: 'homebrewFile',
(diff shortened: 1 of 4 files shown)


stefanbuck:
LGTM 👍  Just one note. I know our code coverage is quite low, but I would like to improve it. Therefore if you could write a test for `homebrew-file.js` that would be a first step in the right direction.

PR #136, opened 2016-07-30
Release 4.2.0

josephfrazier:
See https://github.com/OctoLinker/browser-extension/issues/135

@stefanbuck, as we've discussed privately, I'll take a shot at doing the releases for this one. Just to confirm, here are the steps that are needed (as far as I know):

* [ ] Merge this PR
* [ ] Tag the resulting commit as `v4.2.0` and push the tag into this repo
* [ ] \(Travis automatically releases the Chrome extension, per https://github.com/OctoLinker/browser-extension/commit/2193d36d5892b7f013b64ca18b160963fb9fc926)
* [ ] `npm run package-all`
* [ ] Submit `out/firefox-octolinker-4.2.0.zip` to https://addons.mozilla.org/en-US/developers/addon/octolinker/versions#version-upload
* [ ] Submit `out/opera-octolinker-4.2.0.zip` to https://addons.opera.com/developer/package/226344/?tab=versions

PR #137, opened 2016-07-30
Upon `npm version`: update assets/manifest.json

josephfrazier:
`npm version` is an easy way to commit a new version, but it only
updates `package.json`. This script ensures that the version in
`assets/manifest.json` is kept in sync, to avoid a reoccurrence of
https://github.com/OctoLinker/browser-extension/pull/136#issuecomment-236394248

See here for details about `npm version` and its associated package.json
scripts: https://docs.npmjs.com/cli/version

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "pretest": "npm run lint",
     "test": "karma start",
     "test:watch": "karma start --no-single-run --auto-watch",
+    "version": "json -I -f assets/manifest.json -e \"this.version='`json -f package.json version`'\" && git add assets/manifest.json",
     "package-all": "npm-run-all chrome-pack firefox-pack && cp out/chrome-octolinker-$npm_package_version.zip out/opera-octolinker-$npm_package_version.zip",
     "release": "webstore upload --source out/chrome-octolinker-$npm_package_version.zip --auto-publish",
     "chrome-manifest": "mkdir -p dist && json -e 'delete this.applications; this.permissions.shift()' < assets/manifest.json > dist/manifest.json",


package.json
```diff
@@ -11,6 +11,7 @@
     "pretest": "npm run lint",
     "test": "karma start",
     "test:watch": "karma start --no-single-run --auto-watch",
+    "version": "json -I -f assets/manifest.json -e \"this.version='`json -f package.json version`'\" && git add assets/manifest.json",
```
stefanbuck: Wouldn't do the `git add` in here. This shoudl be handeld by the user I think.

stefanbuck:
Ahh cool, in this case ignore my comment in #136 and keep the version field in the `package.json` file. 👍

package.json
josephfrazier: I think we have to do the `git add` here. I actually forgot it [at first](https://github.com/josephfrazier/browser-extension/commit/3e513acd95c80209bdd488a1ddad0df229507c58), but then the `assets/manifest.json` change wasn't in the generated commit when I tested `npm version patch`. Here's the relevant section of the [documentation](https://docs.npmjs.com/cli/version):

> 4\. Run the `version` script. These scripts have access to the new `version` in package.json (so they can incorporate it into file headers in generated files for example). Again, scripts should explicitly add generated files to the commit using `git add`.

stefanbuck: Ohh I see, thanks for the update.

josephfrazier:
Yeah, I saw your comments in #136 right after opening this :P

At first, I thought that removing the version field from package.json might cause problems, since it's supposedly [required](https://docs.npmjs.com/files/package.json#version), but that doesn't seem to be the case (`npm install` still pulls down the dependencies, for example). However, I think this is a better solution anyway, since it lets us use `npm version`. Thanks for merging it in!

I'll submit a PR for version 4.2.1 shortly.

stefanbuck:
Since there is no new version published in any webstore we can just do the 4.2.0 release again

stefanbuck:
As commented in #136, I removed the release and the git tag from this repo

stefanbuck:
I agree, this solution is much better. Good job

josephfrazier:
Ohhh, ok. Thanks for clarifying. I submitted https://github.com/OctoLinker/browser-extension/pull/138 with a proposal of how to go about the re-release.

PR #140, opened 2016-08-01
Submodule compatibility for npm postinstall symbolic link

### LINKED ISSUE #129: npm install symbolic link failure
~~~
npm ERR! OctoLinker@4.1.0 postinstall: `ln -sf ./scripts/hooks/pre-commit .git/hooks/pre-commit`
npm ERR! Exit status 1
~~~

~~~
~/repos/dotfiles/submodules/chrome/OctoLinker git:master ❯❯❯ ln -sf ./scripts/hooks/pre-commit .git/hooks/pre-commit
ln: .git/hooks/pre-commit: Not a directory
~~~

Freshly added AS SUBMODULE the repo from GitHub, and `.git/hooks` directory does not exist because:

```
~/repos/dotfiles/submodules/chrome/OctoLinker git:master ❯❯❯ cat .git
gitdir: ../../../.git/modules/submodules/chrome/OctoLinker
```

leoj3n:
Closes #129

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   },
   "main": "lib/app.js",
   "scripts": {
-    "postinstall": "ln -sf ./scripts/hooks/pre-commit .git/hooks/pre-commit",
+    "postinstall": "ln -sf `git rev-parse --show-toplevel`/scripts/hooks/pre-commit `git rev-parse --git-dir`/hooks/pre-commit",
     "lint": "eslint .",
     "pretest": "npm run lint",
     "test": "karma start",


stefanbuck:
Hi @leoj3n, I was on vacation so I didn't followed the conversation in #129. I understand you problem, but I'm not sure if it's the right way to solve this like that. In general I'm not that happy with merging code which isn't used in the project. 

What's the reason to have the OctoLinker as a submodule? Why did you need this?

leoj3n:
I like to be able to go in and hack on any Chrome extension I use without losing my settings, and I like the ability to quickly re-provision my extensions to a new machine/account, so I maintain a large list of Chrome extensions as submodules and OctoLinker is [one of them](https://github.com/leoj3n/dotfiles/blob/master/.gitmodules#L138-L141).

I've been trying to think of a better way of doing this that doesn't depend on submodules, so I'd be interested to hear if you have any ideas!

I totally understand not wanting to add unused code, but I honestly didn't expect to run into this error just because I was using the repo as a submodule. I tried using [husky](https://github.com/typicode/husky) to run the `npm run lint` pre-commit hook because I thought this might be an edge case they handled but it errors with 

```
OctoLinker/hooks/prepare-commit-msg: line 3: cd: ../../../../submodules/chrome/OctoLinker: No such file or directory
```

leoj3n:
Looks like they are aware of the problem and have a similar solution proposed:

https://github.com/typicode/husky/issues/20

stefanbuck:
Thanks for the update. One possible solution would be to remove the `postinstall` field from the `package.json` before calling `npm install`. What do you think?

leoj3n:
You mean manually delete the line?

[Skipping the lifecycle scripts](http://stackoverflow.com/a/23506053/906794) like `npm install --ignore-scripts` seems to work fine with the given dependencies, and there are no errors from postinstall because it never gets run.

stefanbuck:
Yes, deleting the line was what I meant. I agree with you, `--ignore-scripts` sounds like a good solution for all of us :fingerscrossed:

Issue #62, opened 2016-02-20
Support for Docker


Gerhut:
```Dockerfile
FROM nodejs:latest
```

https://docs.docker.com/engine/reference/builder/#from

[Docket Hub](https://hub.docker.com/) page would be OK

stefanbuck:
Good idea :+1: As soon as the refactoring is completed (see dev branch), it should be quite easy to implement.

stefanbuck:
Will be part of the next release

PR #141, opened 2016-08-02
Simplify external url handling by remove the ping resolver

stefanbuck:
While working on adding docker support #62, I've noticed that the handling of external urls is a bit nasty. In my opinion a developer shouldn't care about if a url goes to a cross origin page or not. I moved the `live-resolver-ping` parts into the `click-handler` and wrote a couple test around this.

### DIFF
--- a/lib/click-handler.js
+++ b/lib/click-handler.js
@@ -1,6 +1,5 @@
 import $ from 'jquery';
 import { showTooltip } from './gh-interface.js';
-import * as resolvers from './resolver/index.js';
 import { flattenAndCompact } from './utils/array';
 
 const SORRY = 'I\'m sorry, unable to resolve this link  😱';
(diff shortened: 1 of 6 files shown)


josephfrazier:
Wow, nice! I had also tried adding Docker support (I probably should have mentioned it before-hand), but stopped when I ran into problems with CORS (at which point I should have made a "WIP" PR for discussion). To get around them, I tried using [`fetch`](https://devdocs.io/dom/globalfetch) with `mode: 'no-cors'` instead of `$.ajax`, but it turns out that the 'no-cors' mode doesn't even allow access to the response status code, so that didn't work. Using the live-resolver's `/ping` endpoint seems much better.

stefanbuck:
Thanks, it was already there for the nodejs API page so it's not a new feature just better.

Maybe we should start assign issues to avoid that both of us are working on the same task. WIP PR sounds also reasonable. So many progress in here, I'm excited. Thank you so much for your help.

josephfrazier:
Yeah, I wish I had thought to look for `/ping`... It would have saved me a lot of time 😝 

I'll try to remember to assign issues to myself if I'm working on them, that's a good idea. Glad to help!

PR #143, opened 2016-08-03
Publish Chrome extension only when running Node 6

josephfrazier:
During the v4.2.0 release, Travis tried to run the release script on
both Node 4 and Node 6, which resulted in a build [failure]. Besides,
Travis should only try to release from one build environment anyway.

[failure]: https://travis-ci.org/OctoLinker/browser-extension/jobs/148600890#L7346

https://docs.travis-ci.com/user/deployment#Examples-of-Conditional-Releases-using-on%3A

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - 4
   - 6
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


stefanbuck:
Good spot! Maybe it's enough to run the tests on the latest stable version only.

josephfrazier:
Yeah, if you don't have a reason to test on Node v4, I'd prefer to just get rid of it instead. After all, the published extension isn't running under Node anyway.

Issue #144, opened 2016-08-03
#141 broke npm links


josephfrazier:
To reproduce this problem, build the extension from c767c53fed299ded2858f8c9c792c8c5cbc5111a and use it to click on `babel-polyfill` at https://github.com/OctoLinker/browser-extension/blob/730815cc3656b3c96bf4435f97c5ae7eff28a3cb/lib/app.js#L1

There's a TypeError [here](https://github.com/OctoLinker/browser-extension/commit/c767c53fed299ded2858f8c9c792c8c5cbc5111a#diff-34a0c3d0e732cfd9a779519a0ed4c7b0R52) because the function expects to receive a String, but receives a `{url, method}`.

I could probably just fix this myself, but I'm thinking we should probably add some more testing in to make sure this sort of thing can't happen again. @stefanbuck, what do you think?

stefanbuck:
Ohh no. Yes, pleases I'm looking forward to see your fix

PR #145, opened 2016-08-03
Unbreak npm

josephfrazier:
This addresses https://github.com/OctoLinker/browser-extension/issues/144 by reverting the offending commit (https://github.com/OctoLinker/browser-extension/commit/c767c53fed299ded2858f8c9c792c8c5cbc5111a) and using the old way of handling cross-domain links for docker images. I'm thinking that we should merge this, figure out a way to test the correct npm behavior, then find a way to make #141 work again with the new tests.

### DIFF
--- a/lib/click-handler.js
+++ b/lib/click-handler.js
@@ -1,5 +1,6 @@
 import $ from 'jquery';
 import { showTooltip } from './gh-interface.js';
+import * as resolvers from './resolver/index.js';
 import { flattenAndCompact } from './utils/array';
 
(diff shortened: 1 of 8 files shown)


stefanbuck:
I would like to have a look in the actual problem before merging this. I'll keep you posted

josephfrazier:
Sure thing. I briefly tried directly fixing it, as I had mentioned in the other issue, but it ended up not being as straightforward as I expected. You probably have a better idea of how it all fits together, anyway.

PR #148, opened 2016-08-04
homebrew: support `depends_on formula:` syntax

josephfrazier:
Now that https://github.com/OctoLinker/browser-extension/pull/134 is
merged, we can just tweak the regex to support linking to formulas in
the homebrew-core tap. See here for details on the syntax:
https://github.com/caskroom/homebrew-cask/blob/714bd04bebd195500843dbb2cdbef21d65ff8852/doc/cask_language_reference/stanzas/depends_on.md#depends_on-formula

Here's a file for testing:
https://github.com/caskroom/homebrew-cask/blob/42437e5fcd93b1a3c95376794e6d5eee2b096160/test/support/Casks/with-depends-on-formula-multiple.rb#L8


stefanbuck:
👍

PR #149, opened 2016-08-04
Travis: only `npm run package-all` before a deploy

josephfrazier:
Since we don't usually need to have the extension packages, we can save
build time by packaging them only before a deployment, rather than on
every commit. Here's more info on when the various Travis scripts run:
https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ node_js:
 cache:
   directories:
     - node_modules
-after_success:
+before_deploy:
   - npm run package-all
 deploy:
   - provider: releases


stefanbuck:
Good catch. 

The "Update branch" button is again not visible 😕 

screen shot 2016-08-05 at 00 38 02

stefanbuck:
Looks like the following repo setting is the reason for it ... still confused why there is no "update branch" button.

> Require branches to be up to date before merging
> This ensures the branch has been tested with the latest code on master.

PR #152, opened 2016-08-05
Use `export { default as` syntax to dedupe resolver exports

josephfrazier:
This makes it so only one line is needed per resolver.

Maybe we could also do some webpack trickery to make this work like the
plugin auto-loading in ea30b42684cf60d65118562c7ffdcc267067517e, but I
haven't tried to figure that out yet.


stefanbuck:
Nice one

PR #157, opened 2016-08-10
Use https:// in npm-shrinkwrap.json

josephfrazier:


stefanbuck:
How did you do that and why?

josephfrazier:
Oops, I'm sorry if I merged prematurely!

I noticed that when I ran `npm shrinkwrap` on my machine, some of the URLs had changed from http to https, so I figured it'd be best to go ahead and lock in those changes, to take full advantage of https. Note that most of the URLs were already https, so this only changes the remainder.

```sh
git checkout 963cf15
grep https: npm-shrinkwrap.json | wc -l
grep http: npm-shrinkwrap.json | wc -l
```

I'm guessing that when you first ran `npm shrinkwrap`, some of the dependencies had already been downloaded over http, so that got reflected in npm-shrinkwrap.json.

stefanbuck:
Thanks for clearing that up. I did a complete fresh npm install before I was creating the shrinkwrap file. So I  have no clue why some of them are downloaded via http. We will see if the manual editing of the shrinkwrap will cause any further issues as soon we update/add a dependency. By the way, which version of npm do you use? I'm on `3.9.5`. 

Regarding merging. It's fine to merge small stuff by yourself, but new features or bigger/critical PRs I would like to have a discussion first or at lease a thumbs up 😉

josephfrazier:
Hmm, with a fresh `npm install`, it _is_ strange that some dependencies came over http. Hopefully they won't in the future.

For adding/updating dependencies, the [npm documentation](https://docs.npmjs.com/cli/shrinkwrap#building-shrinkwrapped-packages) says that npm-shrinkwrap.json will be updated automatically, as long as we pass `--save` to `npm install`. It looks like `--save-dev` doesn't have the same behavior, but `npm-shrinkwrap.json` doesn't currently have the devDependencies anyway, so that's ok.

I'm on npm `3.9.3` at the moment, so I'm guessing our versions behave mostly the same, according to the [changelog](https://github.com/npm/npm/blob/master/CHANGELOG.md#v395-2016-05-27)

Thanks for the clarification on merging, I'll try to be a little more careful about changing new code like this before getting feedback.

stefanbuck:
> It looks like --save-dev doesn't have the same behavior, but npm-shrinkwrap.json

That's actually a problem for us! I was adding the shrinkwrap especially for them to avoid different build by different versions of babel, webpack .... Will create a new shrinkwrap which includes only the devDep this evening.

Issue #156, opened 2016-08-09
Support gist.github.com?


Qard:
I noticed [gists](http://gist.github.com) don't get the link replacements in require calls.

PR #158, opened 2016-08-10
Add gist support

josephfrazier:
Resolves https://github.com/OctoLinker/browser-extension/issues/156

It turns out that the only thing preventing this from working (other
than permissions) is determining the path associated with the Blob.
Before, the path was undefined, so the Blob was not initialized
properly. With this change, the path will be set to the filename entered
into the gist.

Assuming that that the filename is set to something reasonable, either
the filename-based language detection or the github-based language
detection will work correctly (at least, for non-relative links).

Here's a test page:
https://gist.github.com/josephfrazier/113827963013e98c6196db51cd889c39

---

Note that we may want to change the manifest.json `permissions` handling
at https://github.com/OctoLinker/browser-extension/blob/ec499e065424b5204885624a297ab463c99fa8ab/package.json#L17

See here for prior discussion: https://github.com/OctoLinker/browser-extension/pull/101#issuecomment-227263818

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -36,6 +36,7 @@
   ],
   "permissions": [
     "https://github.com/",
+    "https://gist.github.com/",
     "https://githublinker.herokuapp.com/"
   ]
 }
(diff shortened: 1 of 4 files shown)


stefanbuck:
I had no chance to pick this up .. you are damn fast 🏇  
Good work 👍

PR #159, opened 2016-08-11
Update shrinkwrap to handle devDep

stefanbuck:
Following up on #157, this PR adds only our devDependencies to the shrinkwrap file. 

@josephfrazier the package urls are mixed again with. I don't know why. I performed a `npm cache clean` before, but still the same.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -371,6 +371,48 @@
       "from": "babel-core@>=6.9.1 <7.0.0",
       "resolved": "https://registry.npmjs.org/babel-core/-/babel-core-6.13.2.tgz"
     },
+    "babel-eslint": {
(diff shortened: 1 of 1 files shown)


josephfrazier:
Hmm, what does `npm config get registry` print on your machine? The default value is https://registry.npmjs.org, according to the [docs](https://docs.npmjs.com/misc/config#registry).

At any rate, you can change the URLs to https with this command:

    sed -i 's/http:/https:/g' npm-shrinkwrap.json

LGTM otherwise!

stefanbuck:
Done, but still the behaviour is totally weird to me.

josephfrazier:
Yeah, that's very odd. Maybe reinstalling node/npm would help?

PR #160, opened 2016-08-18
List features 

stefanbuck:
@josephfrazier this PR adds a list of all features that we have in the OctoLinker. I'm not sure if we have to explain every single feature in detail, because it's basically always the same. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,10 @@ Most projects consist of many files and third party dependencies. Files are refe
 
 # Install
 
+<a href="https://chrome.google.com/webstore/detail/octolinker/jlmafbaeoofdegohdhinkhilhclaklkp"><img src="https://raw.githubusercontent.com/alrra/browser-logos/master/chrome/chrome_128x128.png" width="48" /></a>
+<a href="https://addons.mozilla.org/en-US/firefox/addon/octolinker/"><img src="https://raw.githubusercontent.com/alrra/browser-logos/master/firefox/firefox_128x128.png" width="48" /></a>
+<a href="https://addons.opera.com/en/extensions/details/octolinker/"><img src="https://raw.githubusercontent.com/alrra/browser-logos/master/opera/opera_128x128.png" width="48" /></a>
+
 Install OctoLinker from [Chrome Web Store](https://chrome.google.com/webstore/detail/octo-linker/jlmafbaeoofdegohdhinkhilhclaklkp), [Mozilla Add-ons Store](https://addons.mozilla.org/en-US/firefox/addon/octolinker/) or [Opera Add-ons Store](https://addons.opera.com/en/extensions/details/octolinker/).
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,8 +12,82 @@ Most projects consist of many files and third party dependencies. Files are refe
...
+<a href="https://addons.mozilla.org/en-US/firefox/addon/octolinker/"><img src="https://raw.githubusercontent.com/alrra/browser-logos/master/firefox/firefox_128x128.png" width="48" /></a>
+<a href="https://addons.opera.com/en/extensions/details/octolinker/"><img src="https://raw.githubusercontent.com/alrra/browser-logos/master/opera/opera_128x128.png" width="48" /></a>
+
 Install OctoLinker from [Chrome Web Store](https://chrome.google.com/webstore/detail/octo-linker/jlmafbaeoofdegohdhinkhilhclaklkp), [Mozilla Add-ons Store](https://addons.mozilla.org/en-US/firefox/addon/octolinker/) or [Opera Add-ons Store](https://addons.opera.com/en/extensions/details/octolinker/).
 
+# Features
+
+OctoLinker is the easiest and best way to navigate between files and projects on GitHub.com. A click on one of the following keywords will redirect you either to the relative file or to the projects GitHub page. Depending on the value, it may redirects you to an external website like a manual page or another service.
```
josephfrazier: "A click on one of the following keywords ..." isn't completely accurate, but it's tricky to find a better way to describe the functionality. Maybe something like: "Files containing the following keywords will now have links that ..."

"Depending on the value, it may redirects" should say 'redirect' rather than 'redirects'

README.md
```diff
@@ -12,8 +12,82 @@ Most projects consist of many files and third party dependencies. Files are refe
...
+- `conflicts_with`
+- `depends_on cask`
+- `depends_on formula`
+
+### JavaScript
+- `require`
+- `import`
+- `export`
+
+### NodeJS
```
josephfrazier: Assuming the NodeJS list will always be the same as the JavaScript list, maybe we should just combine it into one JavaScript/NodeJS list?

README.md
```diff
@@ -12,8 +12,82 @@ Most projects consist of many files and third party dependencies. Files are refe
...
+- `require`
+- `import`
+- `export`
+
+### npm
```
josephfrazier: I feel like `package.json (npm)` might be a little clearer here.

josephfrazier:
This looks great! I added a few inline comments above, but feel free to merge this as is :)

README.md
stefanbuck: Thanks, I've updated the lines

README.md
stefanbuck: Good point, added it to composer, docker and bower as well

stefanbuck:
@josephfrazier please merge if you are happy with it

PR #162, opened 2016-08-22
Update project website url

stefanbuck:


josephfrazier:
can we do https://octolinker.github.io/ instead?

stefanbuck:
Why should we use https? There are no sensitive data on that page, right?

stefanbuck:
I'm gonna merge this now. We can change it in a later version if needed 😉

josephfrazier:
No problem. It's just that I believe https should be the default for everything if possible. Here's a decent explanation of the reasoning: https://https.cio.gov/everything/

Issue #75, opened 2016-05-16
Support Github Enterprise


ncuillery:
Hello,

Thanks for this awesome chrome extension. It would be great if it could work with Github Enterprise, i.e. the Github interface hosted on a private server (and a private host).

I manage to make it work by

- cloning the repo, 
- add my host in [the manifest](https://github.com/OctoLinker/browser-extension/blob/master/chrome/manifest.json#L15)
- replace `github.com` by `location.host` [here](https://github.com/OctoLinker/browser-extension/blob/master/lib/resolver/javascript-file.js#L30)
- install the extension as an unpacked extension in Chrome.

I don't know if the Chrome extension API allows dynamic content scripts url patterns. Maybe a large permissive pattern is inevitable here 😒

stefanbuck:
Your workaround is indeed not ideal, but in general it's possible. As you already noticed, the code base isn't ready for this feature. Would you like to write a PR for this?

Basicly to make this work, three steps have to be done.

First, an options page have to be added so the user can enter a custom domain. I would recommend to store the value in the LocalStorage. Second, request permissions for the the entered url via [permissions API](https://developer.chrome.com/extensions/permissions#request). Finally, a smallish code refactoring have to be done to get the base url from one place. Check out the  [notifier-for-github-chrome](https://github.com/sindresorhus/notifier-for-github-chrome/) extension to see how this can be implemented. Last but not least, I'll create soon a Firefox version of the OctoLinker, so we have to check how to get this working in both browsers.

I will support you as much as I can to make this happen.

automaticgiant:
just like https://github.com/buunguyen/octotree as another example.

fregante:
I recently published 2 extensions that might help you with this:

```sh
npm i webext-domain-permission-toggle webext-dynamic-content-scripts
```

```js
// in background.js
import 'webext-dynamic-content-scripts';
import addDomainPermissionToggle from 'webext-domain-permission-toggle';

addDomainPermissionToggle();
```

It would also solve https://github.com/OctoLinker/OctoLinker/issues/113

stefanbuck:
This looks amazing @fregante Thanks for sharing

PR #168, opened 2016-08-26
Suporting Github Enterprise with different domains

verychieh:
@stefanbuck Please help to review. This is related to #75 
I tested it with my company github server and worked pretty good.

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -5,6 +5,10 @@
   "author": "Stefan Buck",
   "description": "",
   "homepage_url": "http://octolinker.github.io/",
+  "options_ui": {
+    "page": "options.html",
+    "chrome_style": true
+  },
   "icons": {
     "128": "icon.png"
   },
@@ -21,8 +25,7 @@
   "content_scripts": [
     {
       "matches": [
-        "https://github.com/*",
-        "https://gist.github.com/*"
+        "https://*/*"
       ],
       "js": [
         "app.js"
@@ -35,8 +38,7 @@
     }
   ],
   "permissions": [
-    "https://github.com/",
-    "https://gist.github.com/",
-    "https://githublinker.herokuapp.com/"
+    "storage",
+    "*"
   ]
 }
--- a/lib/resolver/javascript-file.js
+++ b/lib/resolver/javascript-file.js
@@ -6,7 +6,6 @@ export default function ({ path, target }) {
   let basePath = join(dirname(path), target);
   const pathExt = extname(path);
   const fileExt = extname(basePath);
-
   if (fileExt) {
     basePath = basePath.replace(fileExt, '');
   }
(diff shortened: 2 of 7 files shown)


lib/resolver/javascript-file.js
```diff
@@ -27,6 +25,6 @@ export default function ({ path, target }) {
   list.push('');
 
   return list.map((file) => {
-    return 'https://github.com' + basePath + file;
+    return `https://${window.location.hostname}/${basePath + file}`;
```
cehoffman: Perhaps this should be `location.origin` in case they use a non standard port?

assets/manifest.json
```diff
@@ -22,7 +22,8 @@
     {
       "matches": [
         "https://github.com/*",
-        "https://gist.github.com/*"
+        "https://gist.github.com/*",
+        "https://*/*"
```
stefanbuck: The new pattern would match any URL that uses https, which shouldn't be allowed since not every https url is a github instance.

assets/manifest.json
```diff
@@ -37,6 +38,7 @@
   "permissions": [
     "https://github.com/",
     "https://gist.github.com/",
-    "https://githublinker.herokuapp.com/"
+    "https://githublinker.herokuapp.com/",
+    "*"
```
stefanbuck: Changing the permission like that would result in a less restricted extension. The permissions message would change from 

_Read and change your data on gist.github.com, github.com and githublinker.herokupapp.com_

 to 

_Read any change all your data on the websites you visit_

Browser extensions are really powerful and they can do a lot with your browser. The permissions are a good indicator for the user to see how risky an extension could be. Therefore every extensions developer should aim for less as possible restrictions.  

In #75 I mentioned the [optional permission api](https://developer.chrome.com/extensions/permissions) which I think is the best way to deal with additional github enterprise domains.

lib/resolver/javascript-file.js
stefanbuck: Nice workaround to avoid a local storage for the custom domain like described in #75. Anyway, I'm not sure if this still works if the permissions are less loosely again.

stefanbuck:
Thanks @verychieh for the PR. I left some comments where I described my concerns regarding some of your changes. In #75 I described the required parts which have to be done to make this working. 

If you are really interested in adding this functionally, you should start with adding an option page first.

verychieh:
I tried to play around permission in manifest.json but no good luck. So I still have to have "*" in permission field. However, I added some domain checking to avoid loading in octo-linker.js and also options.html to add domains.

stefanbuck:
I think we should split this up into several tasks so that we can work in parallel and review the parts separately.

1. Get base domain (static for now) from a central place
2. Add an empty option page which works in Chrome, Firefox and Opera
3. Add input to option page where user can enter a custom base domain. The domain is stored in LocalStorage (for simplicity)
4. Sort out optional permission on Chrome and add wildcard permission for Firefox and Opera
5. Refactor LocalStorage to browser storage api like chrome.storage

I'll create issues for each of them with some more details. 

Are you happy if we close this PR and start work together on the list above?

verychieh:
okay that is fine with me. I only worked with Chrome and did not consider FF and Opera. It is good to know.

PR #173, opened 2016-08-31
Opening links in new tab on middle mouse click

stefanbuck:
Finally, it's possible to open links in a new tab on middle mouse click 🎉 

dog-home-mfw-EPwELUbhreEPC

### DIFF
--- a/lib/click-handler.js
+++ b/lib/click-handler.js
@@ -71,6 +71,12 @@ function getResolverUrls(dataAttr) {
 }
 
 function onClick(event) {
+  // ignore mouseup events on non-middle buttons
+  // see https://github.com/OctoLinker/browser-extension/pull/173#issuecomment-243895366
+  if (event.type === 'mouseup' && event.which !== 2) {
+    return;
+  }
+
   const dataAttr = event.currentTarget.dataset;
   const $target = $(event.currentTarget);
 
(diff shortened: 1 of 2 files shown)


josephfrazier:
One downside I can think of with this way of solving the problem: if the user tries to select some clickable text (for example, to copy it for searching), the mouseup will cause the browser to navigate away from the page, ruining the user's select operation.

Is there another way to detect middle clicks?

stefanbuck:
Good catch, I didn't think about that. The click event is not working when using the middle mouse on  jQuery delegate element for some reasons. On stackoverflow I found this workaround with mouseup. Let me think about. Anyway excellent spot 👍

josephfrazier:
We might be able to use both a click event and a mouseup event, but ignore the mouseup event if it's not the middle button.

stefanbuck:
Neat idea 👍

stefanbuck:
By adding two event listener one for `click` and another for `mouseup` the `onClick` handler gets called twice for on a normal mouse click. I tried to prevent the mouseup event from bubbling up, but it seems to be impossible http://stackoverflow.com/questions/18342747/why-cant-a-mouseup-event-prevent-a-click-event

A flag which indicates if a click is ongoing could solve the problem, but seems to me a bit dirty to me. Do you have any better ideas?

josephfrazier:
What if we just have `onClick` return immedialy if the `mouseup` event isn't the middle button? I pushed a branch with the changes I have in mind: https://github.com/OctoLinker/browser-extension/compare/5f1f9ef7352557da4db39c59a89269bf40c8eb58...josephfrazier:middle-mouse-click

stefanbuck:
I updated the branch with your suggestion. Looks really good now. Thanks for the support 👍

josephfrazier:
Ooh, with more tests too, nice! I'll go ahead and merge :D

PR #174, opened 2016-08-31
Store base url in a single location

stefanbuck:
This PR replace the hardcoded base url with a placeholder in different places in order to store the base url in a single location. The placeholder will be replaced by the actual base url before a request made.

### DIFF
--- a/lib/resolver/homebrew-file.js
+++ b/lib/resolver/homebrew-file.js
@@ -2,7 +2,7 @@ import relativeFile from './relative-file.js';
 
 export default function ({ path, target }) {
   return [
-    relativeFile({ path, target, baseUrl: 'https://github.com' }),
+    relativeFile({ path, target }),
     'https://github.com/Homebrew/homebrew-core/blob/master/Formula/' + target,
   ];
 }
--- a/lib/resolver/homebrew-file.js
+++ b/lib/resolver/homebrew-file.js
@@ -3,6 +3,6 @@ import relativeFile from './relative-file.js';
 export default function ({ path, target }) {
   return [
     relativeFile({ path, target }),
-    'https://github.com/Homebrew/homebrew-core/blob/master/Formula/' + target,
+    '{BASE_URL}/Homebrew/homebrew-core/blob/master/Formula/' + target,
   ];
 }
--- a/lib/resolver/homebrew-file.js
+++ b/lib/resolver/homebrew-file.js
@@ -2,7 +2,7 @@ import relativeFile from './relative-file.js';
 
 export default function ({ path, target }) {
   return [
-    relativeFile({ path, target }),
-    '{BASE_URL}/Homebrew/homebrew-core/blob/master/Formula/' + target,
+    relativeFile({ path, target, baseUrl: 'https://github.com' }),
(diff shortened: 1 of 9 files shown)


lib/resolver/homebrew-file.js
```diff
@@ -3,6 +3,6 @@ import relativeFile from './relative-file.js';
 export default function ({ path, target }) {
   return [
     relativeFile({ path, target }),
-    'https://github.com/Homebrew/homebrew-core/blob/master/Formula/' + target,
+    '{BASE_URL}/Homebrew/homebrew-core/blob/master/Formula/' + target,
```
josephfrazier: I think this one should stay as https://github.com, since that's the canonical host for Homebrew and there's no guarantee that other hosts would have it mirrored.

stefanbuck: Oh damn, it's too late for coding 😉  Will fix it tomorrow.

stefanbuck: Pushed a new commit where this gets fixed 4d2e7b2e05c5dd98aabb6e1572cf9c316418952b

lib/resolver/homebrew-file.js
```diff
@@ -2,7 +2,7 @@ import relativeFile from './relative-file.js';
 
 export default function ({ path, target }) {
   return [
-    relativeFile({ path, target }),
-    '{BASE_URL}/Homebrew/homebrew-core/blob/master/Formula/' + target,
+    relativeFile({ path, target, baseUrl: 'https://github.com' }),
```
josephfrazier: Oh, I should have explained more in [my previous comment](https://github.com/OctoLinker/browser-extension/pull/174#discussion_r77091612). I don't think `relativeFile` needs to change, only that the fallback URL on the line below this one should always start with https://github.com/Homebrew/homebrew-core. For example, consider the case where you have a custom [Homebrew tap](https://github.com/Homebrew/brew/blob/6622b565ee38de1cd97f5c7b20fa62b481a1ec9b/share/doc/homebrew/brew-tap.md) hosted at https://enterprise-github.example.com/user/repo:

If https://enterprise-github.example.com/user/repo/blob/abcdef/custom-application.rb has a dependency on `custom-library`, it should link to https://enterprise-github.example.com/user/repo/blob/abcdef/custom-library.rb rather than https://github.com/Homebrew/homebrew-core/blob/abcdef/custom-library.rb.

Does that make more sense? If you'd like, I can push up a branch that has my preferred version of this PR.

(See https://github.com/OctoLinker/browser-extension/pull/134 for more context about the homebrew-file resolver)

stefanbuck: Thanks for the explanation, makes sense. Will update the PR later in the day.

stefanbuck: Better now? Please don't merge this PR, because I would like to release a new version tonight. The middle mouse click was a heavily requested feature which I would like to release  independent from this PR.

josephfrazier: Yeah, everything looks good to me :)

I'll leave this and any other PRs alone so the release can go out. Let me know if you'd like any help with it.

stefanbuck: The new release is out

Issue #171, opened 2016-08-30
Add input field for custom domain to options page


stefanbuck:
This issue is a subtask in order to implement #75 and depend on #170.  If you are starting to work on this issue, please leave comment so others are aware of it. Thank you!

This task should add an input field to the option page where the user can enter a custom domain. The field is empty by default. Validation can be ignored for now. The value should be stored in the LocalStorage when the field trigger the blur event.

PR #182, opened 2016-09-15
WIP Add custom domain option field

josephfrazier:
Resolves https://github.com/OctoLinker/browser-extension/issues/171

As seen in the (verbose) commit history, I adapted this code from https://developer.chrome.com/extensions/optionsV2 and tested it in both Chrome and Firefox

Once this looks good to you, @stefanbuck, I'll clean up the commit message upon squash/merging. I wanted to leave it as multiple commits for now to show how the work progressed, as well as to make it easier to tweak anything if needed.

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -41,6 +41,7 @@
   "permissions": [
     "https://github.com/",
     "https://gist.github.com/",
-    "https://githublinker.herokuapp.com/"
+    "https://githublinker.herokuapp.com/",
+    "storage"
   ]
 }
--- a/assets/options.js
+++ b/assets/options.js
@@ -1,0 +1,32 @@
+// Saves options to chrome.storage.sync.
+function save_options() {
+  var color = document.getElementById('color').value;
+  var likesColor = document.getElementById('like').checked;
+  chrome.storage.sync.set({
+    favoriteColor: color,
+    likesColor: likesColor
+  }, function() {
+    // Update status to let user know options were saved.
+    var status = document.getElementById('status');
+    status.textContent = 'Options saved.';
+    setTimeout(function() {
+      status.textContent = '';
+    }, 750);
+  });
+}
+
+// Restores select box and checkbox state using the preferences
+// stored in chrome.storage.
+function restore_options() {
+  // Use default value color = 'red' and likesColor = true.
+  chrome.storage.sync.get({
+    favoriteColor: 'red',
+    likesColor: true
+  }, function(items) {
+    document.getElementById('color').value = items.favoriteColor;
+    document.getElementById('like').checked = items.likesColor;
+  });
+}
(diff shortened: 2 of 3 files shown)


assets/options.js
```diff
@@ -0,0 +1,22 @@
+const customDomainField = document.getElementById('customDomain');
+
+// Saves options to chrome.storage.local.
+function saveOptions() {
+  chrome.storage.local.set({
```
stefanbuck: In general I would prefer `storage.sync` over `storage.local`. However according to https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/storage Firefox and Opera doesn't support `sync`. 

What about using `storage.local` as fallback if `storage.sync` is not available to maintain compatibility.

assets/options.js
```diff
@@ -0,0 +1,22 @@
...
+function restoreOptions() {
+  // Use default value customDomain = ''
+  chrome.storage.local.get({
+    customDomain: '',
+  }, function ({ customDomain }) {
+    customDomainField.value = customDomain;
+  });
+}
+document.addEventListener('DOMContentLoaded', restoreOptions);
+customDomainField.addEventListener('blur', saveOptions);
```
stefanbuck: What happens if the user closes the options dialog by pressing ESC? Will this also trigger a `blur` event?

stefanbuck: btw. should we use babel for this file?

assets/options.js
```diff
@@ -0,0 +1,22 @@
+const customDomainField = document.getElementById('customDomain');
+
+// Saves options to chrome.storage.local.
+function saveOptions() {
+  chrome.storage.local.set({
+    customDomain: customDomainField.value,
+  }, function () {
```
stefanbuck: Does this method require a callback?

stefanbuck: NABD, could be an arrow function 😉

assets/manifest.json
```diff
@@ -41,6 +41,7 @@
   "permissions": [
     "https://github.com/",
     "https://gist.github.com/",
-    "https://githublinker.herokuapp.com/"
+    "https://githublinker.herokuapp.com/",
+    "storage"
```
stefanbuck: We should declare `storage` as an optional permission. Otherwise the update have to be confirmed by the user, because of the new permissions.

josephfrazier:
Thanks for the thorough review! Since this is blocked by https://github.com/OctoLinker/browser-extension/pull/183#pullrequestreview-260574 as you mentioned above, I'll delay making further changes until that's resolved.

stefanbuck:
I closed #183 to clean up the PR list. Are you happy to close this as well? We can reopen it anytime.

josephfrazier:
Yeah, that makes sense :)

Issue #113, opened 2016-06-29
GitLab Support


josephfrazier:
Hey @stefanbuck, I was just thinking I'd like to give this a try, but I'm wondering if you have any feedback on which parts of the code would need to be extended to support it (and of course, if it's something you'd be interested in merging). So far, I've thought of:

* [`blob-reader`](https://github.com/OctoLinker/browser-extension/tree/45f54c5f10b917ffc24d4ac710e0292b442953e8/packages/blob-reader)
* [`github-injection`](https://github.com/OctoLinker/injection)
* [relativeFile resolver](https://github.com/OctoLinker/browser-extension/blob/f30b77bbccd18e4b14e7c990eeb217aaf3090df9/lib/resolver/relative-file.js)
* [rubyFile resolver](https://github.com/OctoLinker/browser-extension/blob/f30b77bbccd18e4b14e7c990eeb217aaf3090df9/lib/resolver/ruby-file.js)
* [javascriptFile resolver](https://github.com/OctoLinker/browser-extension/blob/f30b77bbccd18e4b14e7c990eeb217aaf3090df9/lib/resolver/javascript-file.js)

Here's an example of viewing a Gemfile on the two sites:

* Github: https://github.com/gitlabhq/gitlabhq/blob/49edb2e0f9edf0f717aae78a246efdd23ea08724/Gemfile
* Gitlab: https://gitlab.com/gitlab-org/gitlab-ce/blob/49edb2e0f9edf0f717aae78a246efdd23ea08724/Gemfile

stefanbuck:
Do you think there is a big demand for this? Until now, no one requested GitLab support. If anyone is interested in adding GitLab support, please let us know.

Anyway, in order to add GitLab support we have to touch parts which is also required to support GitHub enterprise #75. Maybe we should start with that first see https://github.com/OctoLinker/browser-extension/issues/75#issuecomment-219690960.

parsingphase:
Bitbucket support also useful - let me know if you prefer a separate ticket.

EricLanduyt:
Just discovered this extension, but at the same time that it is not supporting GitLab, so yes, definitely interested ! :)

yisibl:
I hope we can customize the configuration.

evaryont:
Another request for Gitlab support! Specifically, my own private instance, not just gitlab.com.

tomByrer:
[...]

PR #183, opened 2016-09-15
Add custom domain option to support GitHub Enterprise

stefanbuck:
**This PR is WIP for adding "custom domain" functionality**

Please read the comments for details

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -25,22 +25,22 @@
   "content_scripts": [
     {
       "matches": [
-        "https://github.com/*",
         "https://gist.github.com/*"
       ],
       "js": [
-        "app.js"
-      ],
-      "css": [
-        "style.css"
+        "empty.js"
       ],
       "run_at": "document_idle",
       "all_frames": false
     }
   ],
+  "optional_permissions": [
+    "declarativeContent",
+    "http://*/*",
+    "https://*/*"
+  ],
   "permissions": [
-    "https://github.com/",
-    "https://gist.github.com/",
+    "https://gist.github.com/*",
     "https://githublinker.herokuapp.com/"
   ]
 }
--- a/lib/background.js
+++ b/lib/background.js
@@ -26,3 +26,13 @@ chrome.runtime.onMessage.addListener(({ urls }, sender) => {
     });
   });
 });
+
+// chrome.tabs.onUpdated.addListener((tab, changeInfo) => {
+//   if (changeInfo.status !== 'complete') {
+//     return;
+//   }
+//
+//   chrome.tabs.executeScript({
+//     file: 'app.js',
+//   });
+// });
--- a/lib/options.js
+++ b/lib/options.js
@@ -1,0 +1,26 @@
+document.getElementById('add').addEventListener('click', function () {
+  const url = document.getElementById('url').value;
+
+  chrome.permissions.request({
+    permissions: ['declarativeContent'],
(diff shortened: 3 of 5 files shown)


assets/manifest.json
```diff
@@ -25,22 +25,22 @@
   "content_scripts": [
     {
       "matches": [
-        "https://github.com/*",
```
stefanbuck: I remove the github domain in order to test the functionality. I'll revert this as soon as the PR is ready to go.

assets/manifest.json
```diff
@@ -25,22 +25,22 @@
...
         "https://gist.github.com/*"
       ],
       "js": [
-        "app.js"
-      ],
-      "css": [
-        "style.css"
+        "empty.js"
```
stefanbuck: At least one script must be defined otherwise the extension will be invalid. Therefore I added an empty file. Will revert that to later.

lib/background.js
```diff
@@ -26,3 +26,13 @@ chrome.runtime.onMessage.addListener(({ urls }, sender) => {
     });
   });
 });
+
```
stefanbuck: After obtaining the new host permissions, we have to inject the content script in the matching pages. There are a few ways how to do that:

1. Use the [chrome.declarativeContent.RequestContentScript](https://developer.chrome.com/extensions/declarativeContent#type-RequestContentScript) action to insert a content script in the page.

2. Use the tabs API [chrome.tabs.onUpdated](https://developer.chrome.com/extensions/tabs#event-onUpdated) to detect that a page might have changed, and insert a content script in the page using [chrome.tabs.executeScript](https://developer.chrome.com/extensions/tabs#method-executeScript).

Unfortunately both are not working. 

The first option looks quite promising, but I found a bug which prevent any further https://bugs.chromium.org/p/chromium/issues/detail?id=646650

For the second option we have to ensure that our content script is designed in such a way that it can handle multiple insertions. The babel-polyfill throws an error if the content script is added more than once with `Only one instance of babel-polyfill is allowed`. I didn't found any workaround how to prevent that.

lib/background.js
```diff
@@ -26,3 +26,13 @@ chrome.runtime.onMessage.addListener(({ urls }, sender) => {
...
+// chrome.tabs.onUpdated.addListener((tab, changeInfo) => {
```
stefanbuck: Option two

lib/options.js
```diff
@@ -0,0 +1,26 @@
...
+  chrome.permissions.request({
+    permissions: ['declarativeContent'],
+    origins: [`${url}*`],
+  }, function (granted) {
+    if (!granted) {
+      return;
+    }
+
+    chrome.declarativeContent.onPageChanged.removeRules(undefined, function () {
+      chrome.declarativeContent.onPageChanged.addRules([{
```
stefanbuck: Option one

stefanbuck:
Will close this until there is a chrome fix available

lib/background.js
josephfrazier: The discussion in #113 reminded me of this, and I decided to see how Octotree does it. It looks like they went with the second option, and found a way to avoid inserting the script multiple times: https://github.com/buunguyen/octotree/commit/bdfe39510d005240c4011ca6aa8b5761bce71134#diff-d5d296b6086d9d8de73abd3e81e594cdR4

(I'm not planning to work on this, but it might be useful for anyone who is.)

Issue #163, opened 2016-08-23
Support .NET Core Project.json


rprouse:
The project.json file for .NET Core projects lists dependencies much like NPM. These dependencies tend to link to NuGet.org, but if a nuget.config file is present, it may add or change the package source. I would propose only using NuGet.org as the primary package source since finding the nuget.config file could be problematic.

The reference for project.json is at https://docs.microsoft.com/en-us/dotnet/articles/core/tools/project-json

An example project.json file is,

```json
{
    "version": "1.1.0-*",

    "dependencies": {
        "NUnitWithDotNetCoreRC2": {
            "target": "project"
        },
        "NUnit": "3.4.0",
        "dotnet-test-nunit": "3.4.0-beta-2"
    },
    "testRunner": "nunit",

    "frameworks": {
        "netcoreapp1.0": {
            "imports": [
                "netcoreapp1.0",
                "portable-net45+win8"
            ],
            "dependencies": {
                "Microsoft.NETCore.App": {
                    "version": "1.0.0-*",
                    "type": "platform"
                }
            }
        }
    }
}
```

The nodes of interest are the `dependencies`. Those with `"target": "project"` link to other projects in the current solution/repository and should not be followed.

The link target could be in the form https://www.nuget.org/packages/{PACKAGE}/{VERSION}, for example https://www.nuget.org/packages/dotnet-test-nunit/3.4.0-beta-2

stefanbuck:
Thanks for the detailed description. Would you like to contribute that feature with some guidance and support from our side? If not, no worries I'll put it to our todo list and we will pick it up at some point.

rprouse:
I will take it if I find some time, but I am pretty busy with other projects right now. If someone else wants to pick this up, feel free. I will post a comment if I start on it.

Ibrahim-Islam:
I will start on this

josephfrazier:
Thanks, @Ibrahim-Islam! Let me know if you have any questions.

Ibrahim-Islam:
@josephfrazier I am having some trouble building the project. Can you explain what this [line](https://github.com/OctoLinker/browser-extension/blob/master/package.json#L17) is doing?

stefanbuck:
The `assets/manifest.json` contains values which are Firefox specific. In order to produce a valid `manifest.json` for Chrome this line removes the `applications` value and creates a copy in `/dist/`. 

What issues do you have with building the project? Did you use `npm run chrome-build` to build the extension?

[...]

PR #188, opened 2016-10-09
Use mkdirp package instead of `mkdir -p`

josephfrazier:
This should fix the build on Windows, while still allowing multiple
invocations on Unix-likes, where simply using `mkdir` fails with:

    mkdir: dist: File exists

See https://github.com/OctoLinker/browser-extension/issues/163#issuecomment-252497667


josephfrazier:
@Ibrahim-Islam, could you confirm that this works for you?

stefanbuck:
It works on AppVeyor

PR #189, opened 2016-10-10
Add AppVeyor CI

stefanbuck:
Support Windows CI testing through Appveyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,25 @@
+# Test against this version of Node.js
+environment:
+  nodejs_version: "6"
+
+# Install scripts. (runs after repo cloning)
+install:
+  # Get the latest stable version of Node.js or io.js
+  - ps: Install-Product node $env:nodejs_version
+  # install modules
+  - npm install --no-shrinkwrap --ignore-scripts
+
+# Post-install test scripts.
+test_script:
+  # Output useful info for debugging.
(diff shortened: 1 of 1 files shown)


stefanbuck:
@josephfrazier I added AppVeyor as you suggested in #163. Unfortunately this isn't working yet. I'll have a look at it tomorrow.

josephfrazier:
Hey, I also tried this, and ran into a few problems as well. I've worked around some of them (here's my [work-in-progress](https://github.com/josephfrazier/browser-extension/commits/c067743a4ce3ac20f68a8a0bb81638b6f54840fe)), but there's at least one blocking issue: karma can't seem to load the phantomjs plugin (here is the [build](https://ci.appveyor.com/project/josephfrazier/browser-extension/build/1.0.7)).

stefanbuck:
Thanks for investigating @josephfrazier. I'll update this PR according to your research result.

josephfrazier:
There's a little more to it than that, unfortunately. `npm install` has to be run with `--no-shrinkwrap --ignore-scripts`. For reference, here's the configuration I used at the time of my previous comment. The commit messages have more details about why those flags are required.

https://github.com/josephfrazier/browser-extension/blob/2b854594a079953a3e07dd341c05a5ec6e78c0f5/.appveyor.yml

PR #190, opened 2016-10-11
support for .net core project.json

Ibrahim-Islam:
Issue #163 Few caveats as explained here: https://github.com/OctoLinker/browser-extension/issues/163#issuecomment-252996194


stefanbuck:
I following up on your comment from https://github.com/OctoLinker/browser-extension/issues/163#issuecomment-252996194 here.

I was facing the same issue when I was developing the [npm parser](https://github.com/OctoLinker/browser-extension/blob/0230e594f34e6872dd1343844c2ebbf0370ef030/lib/plugins/npm-manifest/index.js). Mapping the displayed source code values to their related DOM elements is one of the biggest issue that we have. 

I'll try to explain it with a simple JSON example. This is our JSON file and we want to link the first item `hello` from the second array. 

```json
{
  "foo": [
    { "value": "hi" },
    { "value": "hello" }
  ],
  "bar": [
    { "value": "hi" },
    { "value": "hello" }
  ]
}
```

First we have to find the acutely name in the JSON object. In JavaScript you would write something like this:

```js
json.bar[0].value
```

Straightforward and easy to write and read, isn't it? I hear you saying wait a second, we are on GitHub.com there is no such JSON object available ... there is 😉 We have written a library called [blob-reader](https://github.com/OctoLinker/browser-extension/tree/master/packages/blob-reader) which turns GitHubs page source into a unitary structure. A blob has several properties and methods. One of these methods is `.getJSON`. This allows us to treat the blob as a real JSON object. 

Now we know the name of the item, but how to find the related DOM element on page? Luckily there is a fantastic library [findAndReplaceDOMText](https://github.com/padolsey/findAndReplaceDOMText) which searches for regular expression matches in the DOM and replaces or wraps each match with a node or piece of text that you can specify.

Let's do a quick recap. We know the name and with `findAndReplaceDOMText` we can wrap it with a link, so what's the problem? 

The lookup in the DOM to find the corresponding DOM element for our value. In our example above it would link the `hello` from the first array, because the RegExp doesn't know anything about the JSON object structure. Sometimes you can solve this by writing more specific RegExp like I do for `package.json` files. There I build super specific RegExp like `"express" : "v4.0.0"` https://github.com/OctoLinker/browser-extension/blob/0230e594f34e6872dd1343844c2ebbf0370ef030/lib/plugins/npm-manifest/index.js#L8

In you example it even more complicated since the can be nested very deeply. I haven't found a smart solution for this yet. 

Any suggestion or ideas are appreciated.

Ibrahim-Islam:
I think in order for this plugin to support .NET Core `project.json` first it should be decided whether to take the version of package into consideration or not. For dependencies in `package.json` the version is not being taken into account and redirecting to its GitHub page e.g. `https://github.com/babel/babel` rather than `https://github.com/babel/babel/tree/v6.17.0`. So, by that convention, the plugin can do the same for `project.json` and redirect to `https://www.nuget.org/packages/Microsoft.NETCore.App` instead of `https://www.nuget.org/packages/Microsoft.NETCore.App/1.0.0`.

As for the error and in addition to if package version it to be ignored, I can write a `regexBuilder` like yours with only slight change that it will not assume `value` to be a string and ignore if its type of `object`.

@stefanbuck  What do you think?

josephfrazier:
I think it's perfectly acceptable not to take version into account, since that's how other plugins behave :)

> As for the error and assuming if package version is to be ignored, I can write a regexBuilder like yours with only slight change that it will not assume value to be a string and ignore if its type of object.

Instead of writing another `regexBuilder`, you can probably just change [this line](https://github.com/OctoLinker/browser-extension/pull/190/commits/87de270d460d7c0a04d87a259791142612563210#diff-83dfae6ce6b6755eb44b8bf1311f1e00R6) to read:

```js
    const regex = jsonRegExKeyValue(key, String(value));
```

Alternatively, `linkDependency` could simply return early if `value` is an Object.

Ibrahim-Islam:
@josephfrazier Thanks for your comment. I was able to get rid of that exception with your first solution. But since now, version is not a factor then all of the packages below should be linked:
```
  "dependencies": {
    "Microsoft.NETCore.App": {
      "type": "platform",
      "version": "1.0.0"
    },
    "Microsoft.AspNetCore.Diagnostics": "1.0.0",
    "Microsoft.AspNetCore.Mvc": "1.0.0"
  }
```
But `Microsoft.NETCore.App` package will not be linked because from what I understand, `findandreplacedomtext` could not find proper DOM node because it was passed `"[object Object]"`. Is this acceptable?

josephfrazier:
> But Microsoft.NETCore.App package will not be linked because from what I understand, findandreplacedomtext could not find proper DOM node because it was passed "[object Object]". Is this acceptable?

Oh, now I see why it won't work. Personally, I'm ok with a partial solution here, but I think that your idea of writing another `regexBuilder` that ignores the property value is probably better.

stefanbuck:
I agree with @josephfrazier. Even just partial support provides a lot of value to the user. We can improve it over time 😉

Ibrahim-Islam:
If you want I can push the changes for the partial solution? I have never used `findAndReplaceDOMText` before, but in order to get the following match:
```
"Microsoft.NETCore.App": {
   "type": "platform",
    "version": "1.0.0"
  }
```
Should not changing this [line](https://github.com/OctoLinker/browser-extension/blob/master/lib/plugins/helper/regex-builder.js#L5) to following work?
```
const regexValue = escapeRegexString(JSON.stringify(value)); 
```
But it does not work unfortunately.

Ibrahim-Islam:
I have tested the plugin out in `project.json` from the following repos. Seems to be working fine in all cases 😄 

https://github.com/aspnet/cli-samples
https://github.com/aspnet/MusicStore
https://github.com/FabianGosebrink/ASPNET-Core-WebAPI-Sample

lib/plugins/project-json/index.js
```diff
@@ -0,0 +1,23 @@
...
+import { jsonRegExKeyValue } from '../helper/project-json-regex-builder';
+import insertLink from '../../insert-link';
+
+function linkDependency(blob, key, value) {
+  const regex = jsonRegExKeyValue(key, value);
+
+  insertLink(blob.el, regex, {
+    resolver: 'nugetUrl',
+    target: '$1',
+    type: 'project_json',
```
stefanbuck: According to your `nuget-url.js` resolver this property is not needed. 

It's used in the `live-resolver-query` to differentiate between the different registries so our https://github.com/OctoLinker/live-resolver server knows what to do.

lib/plugins/project-json/index.js
```diff
@@ -0,0 +1,23 @@
...
+  });
+}
+
+export default class ProjectJson {
+  getPattern() {
+    return ['project.json'];
+  }
+
+  parseBlob(blob) {
+    processJSON(blob, { '$.dependencies': linkDependency });
```
stefanbuck: Do you think there is a need to support the `tools` field as well? 
https://github.com/aspnet/cli-samples/blob/30932a1738d27bd2567ee6790de9160f2e1e841b/HelloMvc/project.json#L29

Ibrahim-Islam:
@stefanbuck Made the changes as you requested.

stefanbuck:
We will do a release beginning of next week including your changes. Thank you for your first contribution to OctoLinker 👍

Issue #191, opened 2016-10-12
Incorrect display of commented-out code


xwartz:
![image](https://cloud.githubusercontent.com/assets/4149827/19309804/adadf1f8-90b9-11e6-969c-0b85c332ae2a.png)

Live example https://github.com/xwartz/robot/blob/master/src/avatar.js

josephfrazier:
Thanks for the report, @xwartz! I've got a fix pending at https://github.com/OctoLinker/browser-extension/pull/192

stefanbuck:
@xwartz will be fixed in the next release. Thanks @josephfrazier for this hot fix

PR #192, opened 2016-10-12
wrapsInnerString: include text following match

josephfrazier:
Fixes https://github.com/OctoLinker/browser-extension/issues/191


lib/insert-link.js
```diff
@@ -101,9 +101,9 @@ function wrapClosestElement(node, dataAttrObject, matchValue) {
   }
 }
 
-function wrapsInnerString(text, matchValue, dataAttrObject) {
+export function wrapsInnerString(text, matchValue, dataAttrObject) {
```
stefanbuck: I would prefer if we keep this function private. Usually exposing functions just for the purpose of testing is a bad design which ends up most of the times in testing the implementation and not the behaviour. Ideally implementation changes should not affect tests unless the corresponding behaviour is changing as well. Therefor I prefer behaviour over implementation testing.

test/insert-link.test.js
```diff
@@ -127,4 +128,14 @@ describe('helper-replace-keywords', () => {
...
   });
+
+  it('does not remove closing parentheses from commented out require() calls', () => {
+    const text = "// var faker = require('faker')";
+    const options = {
+      'resolver': 'javascriptUniversal',
+      'target': '$1',
+      'path': '/xwartz/robot/blob/3ba4b2628c2252392af58fff0f7231705958ec8d/src/avatar.js',
+    };
+    assert.equal(helper(text, REQUIRE, options).textContent, text);
```
stefanbuck: I think there is no need to pass any options to the helper in this case.

josephfrazier: fixed, thanks 👍

stefanbuck:
Awesome 👍 Regarding publishing a new version, let's wait for #190 since this is just a minor issue.

josephfrazier:
That sounds good, thanks!

PR #201, opened 2016-10-18
helper-grammar-regex-collection: don't capture quotes

josephfrazier:
As we had discussed elsewhere, this is a step towards simplifying
insert-link.js to not worry about quotes.


stefanbuck:
Looks good. Before merging, I would like to do some manual testing first ... just in case.

josephfrazier:
Ok, thanks!

PR #205, opened 2016-10-22
WIP: Test that `insertLink` doesn't rely on quotes

josephfrazier:
As discussed in #201, it'd be nice if `insertLink` didn't have to worry
about quotes. This commit changes `insertLink`'s tests to reflect the
changes made in #201. However, as shown in the test failures,
`insertLink` changes instances of "foo" that aren't in the groups
captured by the regex.

### DIFF
--- a/test/insert-link.test.js
+++ b/test/insert-link.test.js
@@ -4,7 +4,7 @@ import insertLink from '../lib/insert-link.js';
 import { REQUIRE } from '../packages/helper-grammar-regex-collection';
 
 describe('helper-replace-keywords', () => {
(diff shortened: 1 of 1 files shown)


stefanbuck:
👍  looking forward to the final solution

stefanbuck:
@josephfrazier how should we proceed with this PR?

josephfrazier:
I'll close it for now. I never made any more progress on it, and forgot about it, to be honest. At any rate, it doesn't really seem to be a problem, just more of a "nice to have".

PR #206, opened 2016-10-23
Resolve require('module') to https://nodejs.org/api/modules.html

josephfrazier:
Example: https://github.com/kittens/roadrunner/blob/2ead9501e3681fa7584b79e6e6546fb046a83f54/index.js#L3

### DIFF
--- a/lib/resolver/javascript-universal.js
+++ b/lib/resolver/javascript-universal.js
@@ -14,7 +14,11 @@ export default function ({ path, target }) {
   const isBuildIn = builtins.indexOf(target) > -1;
 
   if (isBuildIn) {
-    return `https://nodejs.org/api/${target}.html`;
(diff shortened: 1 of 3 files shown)


lib/resolver/javascript-universal.js
```diff
@@ -14,7 +14,11 @@ export default function ({ path, target }) {
   const isBuildIn = builtins.indexOf(target) > -1;
 
   if (isBuildIn) {
-    return `https://nodejs.org/api/${target}.html`;
+    let filename = target;
+    if (filename === 'module') {
+      filename = 'modules';
+    }
```
stefanbuck: What do you think about wrapping the package `builtins` in a separate file and extend the list there? By that we could easily reuse the list somewhere else, if needed.

PR #210, opened 2016-11-18
python: Try API docs before PyPI

josephfrazier:
Some of the standard library names have been shadowed by user packages
on PyPI. For example, clicking on `argparse` in this [file] should lead
to https://docs.python.org/3/library/argparse.html, not
https://github.com/ThomasWaldmann/argparse/

This commit accomplishes that.

[file]: https://github.com/idank/bashlex/blob/b47fc8a2240af7767ebd33dca39d62a3cf05e0aa/examples/commandsubstitution-remover.py#L3

### DIFF
--- a/lib/resolver/python-universal.js
+++ b/lib/resolver/python-universal.js
@@ -21,10 +21,10 @@ export default function ({ path, target }) {
   }
 
   return [
+    apiDoc,
     liveResolverQuery({
       target: target.split('.')[0],
       type: 'pypi',
     }),
-    apiDoc,
   ];
 }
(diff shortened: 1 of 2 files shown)


stefanbuck:
I stumbled across this and a question poped up. What if an import is referencing to a shadowed package. Is there a way to identify that?

josephfrazier:
Hmm, I guess we could check other files like `requirements.txt` or `setup.py` to see if the pypi package is part of the project, then link to it instead of the API docs, but that seems potentially complicated.

PR #211, opened 2016-11-18
travis: use npm@latest, not 3.10.7

josephfrazier:
(This is a follow-up to https://github.com/OctoLinker/browser-extension/pull/195)

npm@4.0.2 is currently the 'latest' tag, and it has the [fix] for
https://github.com/npm/npm/issues/14042

Note that we can't remove the `before_install` section quite yet,
because no Node.js release currently includes npm@4. On the other hand,
having Travis always use the latest npm might be a nice way to learn
about upcoming breakage...

I also had to regenerate npm-shrinkwrap.json, but it's
backwards-compatible, as shown here:
https://travis-ci.org/josephfrazier/browser-extension/builds/176915380

[fix]: https://github.com/npm/npm/pull/14117

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ node_js:
 cache:
   directories:
     - node_modules
-# TODO unpin npm once this is fixed: https://github.com/npm/npm/issues/14042
+# Use npm > 4 to fix https://github.com/npm/npm/issues/14042
 before_install:
-  - npm install -g npm@3.10.7
+  - npm install -g npm@latest
 before_deploy:
   - npm run package-all
 deploy:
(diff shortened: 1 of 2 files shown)


stefanbuck:
Noob question, it doesn't have an impact on the build result if I do the build locally with npm 3.x and travis with 4.x, right?

josephfrazier:
It shouldn't, but I went ahead and confirmed it by comparing the `unzip` output of the following with both npm@3.10.7 and npm@4.0.2:

```shell
git clean -fxd # WARNING: this removes all files that aren't in git
npm install
npm run chrome-pack
unzip -lv out/chrome-octolinker-4.5.0.zip
```

See here for details: https://superuser.com/questions/578840/comparing-zip-files/578848#578848

PR #216, opened 2016-11-30
Add new window option

stefanbuck:
Add option so user can specify if links should be open in a new tab by default. This is disabled by default.

screen shot 2016-11-30 at 20 05 38

### DIFF
--- a/lib/click-handler.js
+++ b/lib/click-handler.js
@@ -1,6 +1,7 @@
 import $ from 'jquery';
 import { showTooltip } from './gh-interface.js';
 import { flattenAndCompact } from './utils/array';
+import * as storage from './options/storage.js';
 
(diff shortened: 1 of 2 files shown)


lib/click-handler.js
```diff
@@ -100,7 +101,8 @@ function onClick(event) {
 
     showTooltip($target, RESOLVED);
 
-    const newWindow = (event.metaKey || event.ctrlKey || event.which === 2);
+    const newWindow = (storage.get('newWindow') || event.metaKey || event.ctrlKey || event.which === 2);
+    console.log(newWindow, storage.get('newWindow'));
```
josephfrazier: Did you mean to leave this log in?

stefanbuck: No, added that accidentally

Issue #209, opened 2016-11-14
How to communicate features?


stefanbuck:
I have the feeling that a lot of features are not well explained to the users. For example did you know that the linking also works in Pull request? There is a list in the readme which outlines roughly what kind of languages and files are supported, but that's not enough. 

Do you have an idea of how we could do this? I was thinking about videos / images and gifs, but they aren't easy to maintain. Any suggestions are swarml welcome.

stefanbuck:
What about adding a short animation to visualize the link replacement. On the settings page the user can turn off/on this hint.

![](https://cloud.githubusercontent.com/assets/1393946/20439120/74bed740-adba-11e6-826b-98eebd69b395.gif)

dasilvacontin:
I think ideally you shouldn't need to communicate the features. It should just work wherever you'd expect it to work. Knowing the premise, ie 'click to navigate through any import', should suffice.

stefanbuck:
I agree with you @dasilvacontin, but since we don't have full coverage for all languages and every single reference it can be confusing. I think the dot is a good midway solution.

dasilvacontin:
Fair enough. I'll let you know if it ever conflicts with some other extension.

PR #217, opened 2016-11-30
Make links more visible to the user

stefanbuck:
This PR tries to solve #209 by making the octolinker links more visible to the user. I added two versions. 

The first one adds a dot next to the line number. After a short fadeIn animation the dot will stay visible. The other is a bit more pushy. A short background color flash makes the link visible to the user. The user can enable/disable both option separately. 

I'm not sure about the default behaviour. Should we enable both options and point the user to the options page?

### DIFF
--- a/assets/style.css
+++ b/assets/style.css
@@ -1,16 +1,31 @@
-.octo-linker-link {
+.octolinker-link {
   cursor: pointer;
 }
 
-.octo-linker-link:hover {
-  text-decoration: underline;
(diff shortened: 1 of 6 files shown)


stefanbuck:
Thanks for your feedback. Good point with long files. I didn't think about that. I'm also tending to get rid of the flash for the reason that you mention. Also giving user to many options can easily confuse them or lead into adding more and more options. I will ask some colleagues for their opinion on this.

PR #224, opened 2016-12-07
Resolve links in code blocks

stefanbuck:
Finally I managed to resolve links within discussion and other documents like markdown files. By the nature of those files it will not work for relative paths. Also it doesn't work for JSON blocks. Anyway, it's quite handy and provides more value to the user.

## Works in discussions
screen shot 2016-12-08 at 00 47 58

## or in README files
screen shot 2016-12-08 at 00 46 58

### Dependencies

- [x]  #222 needs to be merged before this. FYI @josephfrazier

### DIFF
--- a/assets/style.css
+++ b/assets/style.css
@@ -29,3 +29,12 @@ a.octolinker-link {
     color: rgba(255, 0, 255, .8);
     cursor: default;
 }
+
+/* Fix line indicator for issue code blocks */
+.highlight {
+  position: relative;;
+}
+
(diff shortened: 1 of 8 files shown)


stefanbuck:
@josephfrazier after rebasing the branch is green now. Since you already approved it, I'll merge it by myself.

josephfrazier:
Oops, I didn't consider rebasing at the time. Thanks for fixing it up 👍

PR #232, opened 2016-12-22
Use measurement protocol to get rid of the ga remote script

stefanbuck:
The last release 4.7.0 was rejected by Mozilla, because they disallow `content_security_policy` in `manifest.json` to request remote scripts. We used `content_security_policy` to load https://www.google-analytics.com/analytics.js.

In order to keep the tracking possibility, I implement the [Measurement Protocol](https://developers.google.com/analytics/devguides/collection/protocol/v1/). 

**Dependencies**
- [ ] #234 needs to be merged first


josephfrazier:
> In order to keep the tracking possibility, I implement the Measurement Protocol. Also I split up the background script in smaller parts.

Is it possible to make these two changes in separate pull requests? Having the Measurement Protocol as its own isolated change would make it easier to review/merge/release without having to review the other refactoring as well.

stefanbuck:
Yeah ... it was late and I was tired (bad excuse I know). Anyway I created a new PR for splitting the background script.

lib/background/insight.js
```diff
@@ -1,33 +1,68 @@
...
+      ec: category,
+      ea: action,
+    };
+
+    if (label) params.el = label;
+    if (value) params.ev = value;
+
+    return send({
+      tid: this.tid,
+      cid: this.cid,
```
josephfrazier: Does the Measurement Protocol require a client ID? It would be nice if we can avoid having to track any additional information as a result of this change.

lib/background/insight.js
```diff
@@ -1,33 +1,68 @@
...
+  xhr.send(querystring.stringify(body));
 }
 
-function trackEvent({ category, action, label, value }) {
-  if (!window.ga) {
-    return;
+const send = (requiredParams, params) => {
+  request(baseUrl, {
+    ...params,
+    ...baseParams(requiredParams),
```
josephfrazier: Is the intention here is for the `baseParams` to override the `params`? At first, I thought it might make more sense if the above two lines were reversed in order, but then I realized that I wasn't sure what the goal was.

For reference, here's the documentation I had to re-familiarize myself with: https://babeljs.io/docs/plugins/transform-object-rest-spread/

lib/background/insight.js
stefanbuck: There is no way around, because this is a required property. Before the cid was generated and handled by google analytics itself.

> This anonymously identifies a particular user, device, or browser instance. For the web, this is generally stored as a first-party cookie with a two-year expiration. For mobile apps, this is randomly generated for each particular instance of an application install. 

Our cid is just a random uuid, which we generated once on the first extension launch. There is no reference to a user as you can see [here](https://github.com/octolinker/browser-extension/blob/9e8baf1630f988a7f3b6b7be28e222a0326bbe48/lib/options/storage.js#L23-L26).

lib/background/insight.js
stefanbuck: Yes, the `baseParams` should override the `params`. My indention was to prevent overriding  of important properties like `cid` and `tid` by mistake.

lib/background/insight.js
josephfrazier: Oh, I see now. That seems ok to me, especially since it's anonymous 👍  (from an adversarial perspective, I suppose a future update could de-anonymize any previously collected data, but if there are users worried about that, they're probably already installing the extension from source rather than relying on auto-update)

lib/background/insight.js
josephfrazier: Got it, thanks for confirming.

lib/background/insight.js
stefanbuck: The IP anonymization feature in Analytics sets the last octet of IPv4 user IP addresses and the last 80 bits of IPv6 addresses to zeros. For example `10.215.30.344` will be collected as `10.215.30.0`. By that it's "half" anonymized, because we still know the location of the local network switch, which is good enough.

PR #247, opened 2017-01-10
package.json: Add `chrome-open` script

josephfrazier:
Like the `firefox-open` script, this builds and loads the extension into
an isolated instance of Chrome, using [chrome-launch]. This has a couple
of benefits:

1. New developers can get up-and-running faster
2. It's easier to test Chrome without any other user settings

One caveat so far is that Chrome opens in the background rather than the
foreground, but this is alleviated by printing a message out informing
the developer.

[chrome-launch]: https://github.com/hughsk/chrome-launch

EDIT: Disregard the AppVeyor failure. It's flakiness we've seen before: https://github.com/OctoLinker/browser-extension/pull/115#issuecomment-229819535


package.json
```diff
@@ -48,6 +49,7 @@
     "babel-loader": "6.2.10",
     "babel-plugin-transform-object-rest-spread": "6.20.2",
     "babel-preset-es2015": "6.18.0",
+    "chrome-launch": "1.1.4",
```
stefanbuck: What about declaring `chrome-launch` as an optionalDependencies along with `web-ext`?

josephfrazier: Hmm, I'm not sure if this is a good idea. I tried two different ways of doing it:

1. Move those packages from `devDependencies` to `optionalDependencies`. However, due to the shrinkwrap, this results in those packages not getting installed at all. Here's a commit where I tried it, with more details in the message: https://github.com/josephfrazier/browser-extension/commit/8e18566add013ff8e52c7fe14650e494f1f55d35

2. Just add those packages to `optionalDependencies`, while keeping them in `devDependencies` and npm-shrinkwrap.json. I'm not sure if this accomplishes anything or even makes sense, given https://github.com/npm/npm/issues/3870. It doesn't seem to break the install/chrome-open workflow, for what it's worth. Here's the test commit: https://github.com/josephfrazier/browser-extension/commit/95b6e666df5dbe1dd9e5c54799603232e502b424

What do you think?

stefanbuck: Ahhh I wasn't aware of the issue with the shrinkwrap. Declaring the dependencies twice doesn't sound right either. Let's continue with the good old way 😉

josephfrazier: Awesome, thanks for merging!

PR #251, opened 2017-01-20
Rename methods to be more consistent with the js standard

stefanbuck:


packages/blob-reader/blob.js
```diff
@@ -7,13 +7,13 @@ export default class Blob {
...
-  getText() {
+  toString() {
     return this.lines.map(({ value }) => value).join('\n');
   }
 
-  getJSON() {
+  toJSON() {
     try {
-      return JSON.parse(this.getText());
+      return JSON.parse(this.toString());
```
stefanbuck: I know, `toString()` isn't needed here, but it's more explicit.

Issue #250, opened 2017-01-20
Need help with RegExp for Go import statements 


stefanbuck:
I started working on adding support for Go #46. The beginning was quite simple, since the import paths are often just a repository url like `import "github.com/spf13/cobra`. I'm aware of [custom import paths](https://texlution.com/post/golang-canonical-import-paths/), which I'll tackle in a later PR.

Back to my issue. Go has two kind of import statements, singe and multiple.

### Single import

```go
import "path"
import "github.com/spf13/cobra`
```

### Multiple imports
```go
import (
  "path"
  "github.com/spf13/cobra`
)
```

As you can see, the single import is similar to JavaScript version and easy to implement with a RegExp. I'm stuck with RegExp for the multiple import. My best result is `import \((?:[\s\n])+"([^"\s]+)"]*` which grabs only the first import value. We need some kind of loop or precondition ... maybe it's even impossible with a RegExp. 

Then I was thinking about writing an AST parser for Go. I never build an AST parser before, but after reading more about it, it sounds way too complex just for our purpose of identifying import values.

Another idea would be to build a sequential line parser which excepts three RexExp. The first one defines the "opening" state, the second one does the actually matching and the last one defines the end ... roughly explained. I can follow up on this if needed.

At the end we still need a RegExp for wrapping the value with a link, because our wrapping library [findAndReplaceDOMText](https://github.com/OctoLinker/browser-extension/blob/4c78d4ad64ee7094b9cc5f4b381076dd4e35bfe3/lib/insert-link.js#L173) can only work with RegExp.

That's my ideas ... maybe someone has a better / simpler solution.

stefanbuck:
Quick and dirty line parser so you get a rough understanding of the idea  https://github.com/stefanbuck/browser-extension/blob/51dc9ba6ca143ab69a4b26d852279a8853b15e83/lib/plugins/go/index.js#L7-L47

Issue #248, opened 2017-01-18
Unhandled promise rejection in storage.js


josephfrazier:
One of the Firefox extension reviewers left this comment about version 4.7.1:

> I'm getting an error with this add-on installed, since Firefox does not yet have chrome.storage.sync enabled. For the next version, can you make sure to fall back to chrome.storage.local if the sync methods reject the promise/throw an error?

> 18:02:55.092 Unhandled promise rejection Error: Please set webextensions.storage.sync.enabled to true in about:config

I haven't yet reproduced the problem myself, as I primarily use Chrome, but the relevant code appears to be on [lines 14 and 20 in storage.js](https://github.com/OctoLinker/browser-extension/blob/0db32d949b110152f3689fab183c534272c97efb/lib/options/storage.js#L14-L20). As suggested in the above comment, it should be pretty straightforward to add a more comprehensive fallback to `chrome.storage.local`.

We should probably also consider handling the case where `chrome.storage.local` fails as well (though I'm not sure if this happens in practice). Perhaps a `storage` API wrapper library would be helpful here.

stefanbuck:
A wrapper around the storage sounds like a good idea :+1:

josephfrazier:
I was able to reproduce the issue by running:

```shell
npm run firefox-open -- --firefox=firefoxdeveloperedition
```

and navigating to https://github.com/OctoLinker/browser-extension/blob/master/package.json

I'll keep looking into it.

stefanbuck:
Another option would be to use [rfpify](https://www.npmjs.com/package/rfpify) to promisify the chrome extension api directly.

josephfrazier:
`rfpify` looks pretty nifty, but the `chrome.storage` functions have an additional complication of putting their errors into `chrome.runtime.lastError`, so I'm hoping to find a wrapper than can properly deal with that. I've pushed up some work-in-progress to https://github.com/OctoLinker/browser-extension/pull/258

PR #253, opened 2017-01-23
Add support fo golang

stefanbuck:
With this PR, Go is finally supported by OctoLinker.

go

### Core library imports

A standard library import like `import "math/big"` will reference to `https://golang.org/pkg/math/big`.

### Relative imports

An import path beginning with `./` or `../` is treated as a relative path. OctoLinker tries to resolve the path and redirects you to the linked resource.

### Remote imports

Last but not least, OctoLinker also supports hosting sites. For example `import "github.com/foo/bar"` will reference to `https://github.com/foo/bar`. The following hosting sites are supported:

- github.com
- bitbucket.org
- launchpad.net
- hub.jazz.net
- gopkg.in

**Note, other remote import path like `company.com/foo/bar` aren't supported yet.** If you want to get this implemented vote on the related issue #264.

---

# Technical details

I marked this PR as `to discuss`, because it introduce some new techniques, which enables to process complex syntax easily. 

As mention in #250, the multi line import statements aren't easy to catch with a RegExp. Therefore I added a "[line by line](https://github.com/OctoLinker/browser-extension/compare/master...stefanbuck:go?expand=1#diff-03f14f22ff89ce935b0f02aa38909fb7)" parser. This parser reads the given sourceString line by line. The parser expects three RegExp `openingPattern` `closingPattern` and `matchPattern`. The `openingPattern` determines the beginning of the "crawling" process. Every line followed by the `openingPattern`, gets evaluated by `matchPattern` until the end is reached by matching `closingPattern`. The result of the `matchPattern` will be returned. 

```go
package foo   // This line will be ignored
import (      // Opening pattern
  "foo"       // match
  "bar"       // match 
)             // Closing pattern
              // This line will be ignored
```

In this case the return values will be
```js
[
  ['foo'],
  ['bar']
]
```

Note, it's an array of arrays, because an RegExp group can contain more than one value. 

We are half way through, let's crack one. Based one the returned values from the parser, we build a super specific RegExp for every single match. That's the reason why `packages/helper-grammar-regex-collection/index.js` exposes a function instead of a RegExp for `go`. The return value of this function looks like

```js
[
  /import\s\([^]+"(foo)"/,
  /import\s\([^]+"(bar)"/,
  /import\s['"]([^'"\s]+)['"]?/g
]
```

The third line is there to match also single line imports. In the go plugin itself, we loop over this array to invoke the link injection. That's it. 

Does it make sense?


josephfrazier:
Thanks for the detailed explanation! I saw your comments in #250, but hadn't gotten a chance to respond to them. I think your line-by-line approach here is a good alternative to a full-fledged AST parser, and the rest of the implementation looks good to me as well 👍 (and I agree that this seems impossible to do with a single RegExp)

I'd be happy to merge these changes once the commit message is improved and tests have been added. I can also help out with those things, if you'd like. Cheers!

stefanbuck:
Sure, I'll rephrase/finalise the commit message and add some more tests. I don't think I need help with this, anyway thanks for asking. It would be great if you could work on #248 so we can solve this in the next release as well.

stefanbuck:
It's almost complete. There is only one issue left to resolve before we can merge this feature. As mentioned earlier, we don't support custom import path like `code.gitea.io/gitea/modules/log`. Therefore I would like to not convert those strings into links.

josephfrazier:
Oh, I had forgotten about [those](https://golang.org/cmd/go/#hdr-Remote_import_paths) (linked for future reference). I look forward to seeing the result!

packages/helper-grammar-regex-collection/test.js
```diff
@@ -260,7 +261,7 @@ describe('helper-grammar-regex-collection', () => {
     });
   });
 
-  describe('go()', () => {
+  describe.only('go()', () => {
```
josephfrazier: It looks like this made it into the commit by accident :)

stefanbuck:
@josephfrazier this is ready for review 😉

packages/helper-grammar-regex-collection/test.js
stefanbuck: 😊

packages/byline-parser/index.js
```diff
@@ -34,7 +34,8 @@ export default function (blobSource, { openingPattern, closingPattern, matchPatt
         throw new Error('No capture group is defined');
       }
 
-      tmp.push(captureGroup.slice(1));
+      if (!noGroup) tmp.push(captureGroup.slice(1));
+      if (noGroup) results.push(captureGroup.slice(1));
```
stefanbuck: I'm not happy with this solution. I tried to avoid code duplication, but now I've the feeling this will be horrible to maintain.

josephfrazier: Do the opening and closing pattern need to be optional for Go? I'm wondering if we can just get rid of this commit entirely. Of course, even if Go doesn't need this functionality, it might be useful later.

In any case, these two lines could be combined into:

```js
(noGroup ? results : tmp).push(captureGroup.slice(1));
```

stefanbuck: Are you aware of that, go supports two syntax definitions for import?

### Single import
```go
import "os";
import "github.com/foo/bar";
```

### Multi imports
```go
import (
  "os"
  "github.com/foo/bar"
)
```

When I started with this PR, the single import was handled with a RegExp, like we do for other languages. This will match every import regardless of the value. 
While writing this PR, we don't support custom import paths and therefore I was thinking it would be better to prevent the link injection in the first place to avoid confusions. 

With 8f7cad0ddec4ea0fa38360a1625e084b94fe366e I restricted it for multi imports already. To implement that same behaviour for single imports, we have to use the RegExp builder like we do for mutli imports. 

Sorry, I should have explained it better, earlier. Does it make sense now?

josephfrazier: Oh, I see now. I was aware of the multiple import syntaxes, but I didn't see the need to omit the opening/closing patterns in order to ensure that `import "custom.tld/path"` isn't linked. To me, it looked like the intent was to match text like:

```golang
"os"
"github.com/foo/bar"
```

that simply doesn't have the `import (` or the `)`. It all makes sense now, though. Thanks for the clarification!

r-obert:
:tada:

PR #257, opened 2017-01-25
package.json: Extract firefox-launch script

josephfrazier:
package.json: Extract `firefox-launch` script

This allows developers to open the extension in Firefox without
rebuilding the extension, which is useful if they have `npm run
firefox-watch` running in another shell.

The `firefox-open` script has `--` on the end so that arguments can
still be passed through it to `web-ext`, as in:

    npm run firefox-open -- --firefox=firefoxdeveloperedition

(see https://github.com/OctoLinker/browser-extension/issues/248#issuecomment-273624297)


stefanbuck:
Good idea 👍

PR #260, opened 2017-01-26
.appveyor.yml: Retry Chocolatey installs and PhantomJS

josephfrazier:
This adds an `appveyor-retry` wrapper around parts of the AppVeyor build
that have intermittently failed in the past. Here are some example
failures:

* https://ci.appveyor.com/project/josephfrazier/browser-extension/build/1.0.44
* https://ci.appveyor.com/project/stefanbuck/browser-extension/build/1.0.177

See the commit messages for more details.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,8 +9,8 @@ cache:
 install:
   # Get the latest stable version of Node.js or io.js
   - ps: Install-Product node $env:nodejs_version
-  - choco install phantomjs
-  - choco install zip
+  - appveyor-retry choco install phantomjs
+  - appveyor-retry choco install zip
   # install modules
   - npm install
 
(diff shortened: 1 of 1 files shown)


stefanbuck:
Good job 👍 Maybe we can move away from PhantomJS and replace it with Headless Chromium someday.

josephfrazier:
Another potential alternative is [jsdom](https://github.com/tmpvar/jsdom), assuming the tests only need to know about the DOM, rather than requiring layout/rendering or other "real browser" stuff. See here for more details on the differences between PhantomJS and jsdom: https://github.com/tmpvar/jsdom#jsdom-vs-phantomjs

stefanbuck:
The other day, I was thinking about using a headline chrome instance for testing. I know jsdom and I really like it. The only downside is that jsdom doesn't support web extension API's. 

Right now, we use a [fake api](https://github.com/OctoLinker/browser-extension/blob/80cf375a2c1f65a204732b25d20007a684c7f005/test/main-lib.js#L3-L10) implementation in order to test the API, but this feels a bit messy.

PR #263, opened 2017-01-28
Remove pre-commit hook

josephfrazier:
I often use commits to save work-in-progress, with the intention of
revising them later. Removing the pre-commit hook makes this faster.

Since `npm run lint` is also in the `pretest` script, it will still be
run on `npm test`, which should be done anyway before opening a pull
request.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -183,7 +183,7 @@
         "glob": {
           "version": "7.1.1",
           "from": "glob@>=7.0.0 <8.0.0",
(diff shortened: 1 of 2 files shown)


stefanbuck:
Fair point

josephfrazier:
Thanks! Until recently, it was easy enough for me to just delete `./git/hooks/pre-commit` manually. However, when I started working on https://github.com/OctoLinker/browser-extension/issues/248, and `npm install`-ing different sets of packages, `husky` would always bring the hook back.

PR #266, opened 2017-01-30
Fix go package imports

stefanbuck:
While testing the go implementation, I've noticed that some imports doesn't work. I thought that go  reference repo like `github.com/foo/bar` only. I didn't know that you can also import packages form your project like `github.com/foo/bar/baz`. This should point to `https://github.com/foo/bar/tree/master/baz` or `https://github.com/foo/bar/blob/master/baz/baz.go`.

### DIFF
--- a/lib/resolver/go-universal.js
+++ b/lib/resolver/go-universal.js
@@ -1,10 +1,34 @@
 import goFile from './go-file.js';
 
+function githubUrls(url) {
+  const [, user, repo, ...path] = url.split('/');
+
+  if (!path.length) {
+    return [
+      `{BASE_URL}/${user}/${repo}/blob/master/${repo}.go`,
+      `{BASE_URL}/${user}/${repo}`,
+    ];
+  }
+
+  const fullPath = path.join('/');
+  const last = path.slice(-1);
+
+  return [
+    `{BASE_URL}/${user}/${repo}/blob/master/${fullPath}/${last}.go`,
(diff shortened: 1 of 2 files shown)


test/resolver/go-universal.test.js
```diff
@@ -0,0 +1,48 @@
+import assert from 'assert';
+import goUniversal from '../../lib/resolver/go-universal.js';
+
+describe.only('go-universal', () => {
```
josephfrazier: I think this was supposed to be:

```js
describe('go-universal', () => {
```

lib/resolver/go-universal.js
```diff
@@ -1,12 +1,36 @@
...
+}
+
 export default function ({ target, path }) {
   const isPath = !!target.match(/^\.\.?[\\|\/]?/);
 
   if (isPath) {
     return goFile({ path, target });
   }
 
+  if (target.startsWith('github.com')) {
```
josephfrazier: If we ever [support GitHub Enterprise](https://github.com/OctoLinker/browser-extension/issues/75), we'll have to change this into a more general `isGithubUrl` function (similar to what happened in 71ae1d68526919bee2c6e8339a1cd87b5febff11). I think it's ok for now though, especially since it's currently the only instance of this logic. 👍

josephfrazier:
Other than the `describe.only` mentioned [above](https://github.com/OctoLinker/browser-extension/pull/266#discussion_r98565245), this looks great!

Issue #271, opened 2017-02-01
Cannot resolve module names containing dots 


josephfrazier:
For example, `./modules/es6.symbol` at https://github.com/zloirock/core-js/blob/0bb657cfc7834d03fd68c48a1cba70f2ccab20c0/shim.js#L1 should resolve to https://github.com/zloirock/core-js/blob/0bb657cfc7834d03fd68c48a1cba70f2ccab20c0/modules/es6.symbol.js, but it doesn't.

PR #273, opened 2017-02-01
javascript-file: Don't strip extension from file names

josephfrazier:
fixes https://github.com/OctoLinker/browser-extension/issues/271

This allows paths like `./modules/es6.symbol` to correctly resolve to
`./modules/es6.symbol.js`

For example: https://github.com/zloirock/core-js/blob/0bb657cfc7834d03fd68c48a1cba70f2ccab20c0/shim.js#L1

EDIT:

Removing the `fileExt` code didn't break any tests, but I'm not sure if it breaks any functionality in practice. @stefanbuck, do you know why it was there?

### DIFF
--- a/lib/resolver/javascript-file.js
+++ b/lib/resolver/javascript-file.js
@@ -3,20 +3,11 @@ import { join, dirname, extname } from 'path';
 export default function ({ path, target }) {
   const list = [];
   const extName = ['.js', '.jsx', '.ts', '.tsx', '.json'];
-  let basePath = join(dirname(path), target);
+  const basePath = join(dirname(path), target);
   const pathExt = extname(path);
-  const fileExt = extname(basePath);
-
-  if (fileExt) {
-    basePath = basePath.replace(fileExt, '');
-  }
 
   if (pathExt && !extName.includes(pathExt)) {
     extName.unshift(pathExt);
-  }
-
-  if (fileExt && !extName.includes(fileExt)) {
-    extName.unshift(fileExt);
   }
 
   extName.forEach((ext) => {
--- a/lib/resolver/javascript-file.js
+++ b/lib/resolver/javascript-file.js
@@ -1,10 +1,18 @@
 import { join, dirname, extname } from 'path';
 
-export default function ({ path, target }) {
+function resolve({ path, target }) {
   const list = [];
   const extName = ['.js', '.jsx', '.ts', '.tsx', '.json'];
   const basePath = join(dirname(path), target);
   const pathExt = extname(path);
+  const fileExt = extname(basePath);
+
+  if (extName.includes(fileExt)) {
+    return resolve({
+      path,
+      target: target.replace(fileExt, ''),
+    });
+  }
 
   if (pathExt && !extName.includes(pathExt)) {
     extName.unshift(pathExt);
@@ -19,3 +27,5 @@ export default function ({ path, target }) {
 
   return list.map(file => `{BASE_URL}${basePath}${file}`);
 }
+
+export default resolve;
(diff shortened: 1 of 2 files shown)


stefanbuck:
Not from top of my head. I'll have a look later today

stefanbuck:
An issue occurs when the path contains a file extension. Before `./modules/es6.symbol.js` was resolved to

```
{BASE_URL}modules/es6.symbol.js
{BASE_URL}modules/es6.symbol/index.js
{BASE_URL}modules/es6.symbol.jsx
{BASE_URL}modules/es6.symbol/index.jsx
{BASE_URL}modules/es6.symbol.ts
{BASE_URL}modules/es6.symbol/index.ts
{BASE_URL}modules/es6.symbol.tsx
{BASE_URL}modules/es6.symbol/index.tsx
{BASE_URL}modules/es6.symbol.json
{BASE_URL}modules/es6.symbol/index.json
{BASE_URL}modules/es6.symbol
```

Every line is invalid here, except the fist one. Now it's the opposite, only the last line is valid. This results in many unnecessary requests which leads to loading delays.

```
{BASE_URL}modules/es6.symbol.js.js
{BASE_URL}modules/es6.symbol.js/index.js
{BASE_URL}modules/es6.symbol.js.jsx
{BASE_URL}modules/es6.symbol.js/index.jsx
{BASE_URL}modules/es6.symbol.js.ts
{BASE_URL}modules/es6.symbol.js/index.ts
{BASE_URL}modules/es6.symbol.js.tsx
{BASE_URL}modules/es6.symbol.js/index.tsx
{BASE_URL}modules/es6.symbol.js.json
{BASE_URL}modules/es6.symbol.js/index.json
{BASE_URL}modules/es6.symbol.js
```

josephfrazier:
Hmm, I see what you're saying. That is a bit tricky... What if we assume that paths ending in a js/ts/json extension are paths to files, rather than paths to folders? Of course, a reference to a _folder_ named `./modules/es6.symbol.js` would still fail to be correctly resolved to `{BASE_URL}modules/es6.symbol.js/index.js`, but this seems like a much smaller edge case. I think people are less likely to end folder names with `.js` (or similar extensions) than they are to put a dot in a filename.

I'll push up a couple of commits that implement this approach. Let me know what you think.

stefanbuck:
I agree with you, that it's less likely to end folder names with file extensions, but there is another issue. A folder could also contain a dot as a delimiter like `config.dev`. Since `path.extname()` doesn't know if the last portion  is a "real" file extension or just a string this wouldn't work. 

> The path.extname() method returns the extension of the path, from the last occurrence of the . (period) character to end of string in the last portion of the path. 

_[https://nodejs.org/api/path.html#path_path_extname_path](https://nodejs.org/api/path.html#path_path_extname_path)_

The behaviour after your refactoring was correct. Just the fact that the most common case was at the end of the array was worrying me.

josephfrazier:
> Since path.extname() doesn't know if the last portion is a "real" file extension or just a string this wouldn't work.

The current changes make it so that only the `['.js', '.jsx', '.ts', '.tsx', '.json']` extensions are considered to be "real" file extensions, so I don't think this is a big problem: https://github.com/OctoLinker/browser-extension/pull/273/commits/e19c030cf89d5b46b222d9c56462ca175adb649b#diff-e822edbd04f428360d5ab5038457a35fR10

---

Alternatively, I could revert the two WIP commits to go back to the initial refactoring, but then add a condition that tries the "extension-less" path first if the path already ends in a  `['.js', '.jsx', '.ts', '.tsx', '.json']` extension. What do you think?

stefanbuck:
I haven't seen the new commits, because they didn't appeared after you comment 😊 Anyway, that looks pretty good now 👍

josephfrazier:
Oops, I didn't consider that! In the future, I'll try to push changes before making comments. Thanks for the merge!

P.S. If you'd like to get emails when commits are added in a pull request, you can use the **Pull Request pushes** checkbox here: https://github.com/settings/notifications

stefanbuck:
No worries, it was my fault. Next time I'll look twice.

PR #274, opened 2017-02-01
Click handler refactoring proposal

stefanbuck:
Once @josephfrazier suggested to combine the plugin click handler (aka resolver) with the actual plugin. At this time, I was assuming that we can benefit from reusing those resolver from one plugin in others. Now, after a couple of months of adding more and more languages, it shows up that only a small portion ( [git-url](https://github.com/OctoLinker/browser-extension/blob/e23ab38e07fb53128e3370a9d5c66734ffa5a4c6/lib/resolver/git-url.js), [relative-file](https://github.com/OctoLinker/browser-extension/blob/e23ab38e07fb53128e3370a9d5c66734ffa5a4c6/lib/resolver/relative-file.js) and [live-resolver-query](https://github.com/OctoLinker/browser-extension/blob/e23ab38e07fb53128e3370a9d5c66734ffa5a4c6/lib/resolver/live-resolver-query.js)) of resolver are reused.

Exemplary I refactored the Docker and Python plugin to show you how it could work. I'll polish the PR if you agree with the solution. I'm looking forward to your feedback.

Please, review the PR with [?w=1](https://github.com/OctoLinker/browser-extension/pull/274/files?w=1) to ignore whitespace changes.

### TODO
- [x] Refactor Docker plugin
- [x] Refactor Python plugin
- [x] Refactor Rust plugin
- [x] Refactor Vim plugin
- [x] Refactor Ruby plugin
- [x] Refactor Go plugin
- [x] Refactor .Net Core plugin
- [x] Refactor Rubygems plugin
- [x] Refactor Homebrew plugin
- [x] Refactor Composer plugin
- [x] Refactor JavaScript plugin
- [x] Refactor TypeScript plugin
- [x] Refactor npm plugin
- [x] Refactor Bower plugin


josephfrazier:
This is a big one, so I'll try to take a look at it over the weekend. Thanks for the heads-up about the whitespace changes.

stefanbuck:
No rush :wink:

lib/insert-link.js
```diff
@@ -54,8 +54,11 @@ function getCaptureGroupValue(match, captureGroup) {
   return captureGroup.replace(new RegExp(`\\$${index}`, 'g'), match[index]);
 }
 
-function buildDataAttr(data, match) {
-  const dataAttr = {};
+function buildDataAttr(_pluginName, data, match) {
+  const dataAttr = {
+    _pluginName,
```
josephfrazier: EDIT: See this comment first: https://github.com/OctoLinker/browser-extension/pull/274/files#r99663118

If you agree with my comment at https://github.com/OctoLinker/browser-extension/pull/274/files#r99657441, we should also change `_pluginName` to `pluginName` on this  line (and in the function definition).

lib/click-handler.js
```diff
@@ -50,47 +49,45 @@ function tryLoad(urls, cb) {
...
-        }
-
-        return {
-          url: `https://githublinker.herokuapp.com/ping?url=${url}`,
-          method: 'GET',
-        };
-      });
-    }
+  const pluginName = dataAttr._pluginName;
+  const resolver = pluginManager.getClickHandler(pluginName);
```
josephfrazier: I think we should rename `getClickHandler` to `getResolver`, since that more accurately describes what the return value does (and matches the variable name used here).

lib/plugin-manager.js
```diff
@@ -53,9 +53,26 @@ function getPluginsForGithubClasses(plugins, classList) {
 
 export default class {
   constructor(plugins) {
+    this._pluginsList = plugins;
     this._plugins = buildPluginCache(plugins);
   }
 
+  getClickHandler(pluginName) {
```
josephfrazier: If you agree with my comment at https://github.com/OctoLinker/browser-extension/pull/274/files#r99658228, we should also change `getClickHandler` to `getResolver` here.

lib/plugin-manager.js
```diff
@@ -53,9 +53,26 @@ function getPluginsForGithubClasses(plugins, classList) {
...
+      console.error(`Plugin ${pluginName} not found`);  // eslint-disable-line no-console
+      return;
+    }
+
+    if (!plugin.clickHandler) {
+      console.error(`No clickHandler found for ${pluginName}`);  // eslint-disable-line no-console
+      return;
+    }
+
+    return plugin.clickHandler;
```
josephfrazier: In the spirit of https://github.com/OctoLinker/browser-extension/pull/274/files#r99658228, I think `clickHandler` should instead be `resolve` on this line and in the previous conditional.

lib/plugins/docker/index.js
```diff
@@ -4,13 +4,25 @@ import preset from '../../pattern-preset';
 
 export default class Docker {
 
+  static clickHandler({ image }) {
```
josephfrazier: If you agree with my comment at https://github.com/OctoLinker/browser-extension/pull/274/files#r99658736, we should also change `clickHandler` to `resolve` here.

lib/plugins/python/index.js
```diff
@@ -1,16 +1,45 @@
 import { PYTHON_IMPORT } from '../../../packages/helper-grammar-regex-collection/index.js';
 import insertLink from '../../insert-link';
 import preset from '../../pattern-preset';
+import liveResolverQuery from '../../resolver/live-resolver-query.js';
+import relativeFile from '../../resolver/relative-file.js';
 
 export default class Python {
 
+  static clickHandler({ path, target }) {
```
josephfrazier: If you agree with my comment at https://github.com/OctoLinker/browser-extension/pull/274/files#r99658736, we should also change `clickHandler` to `resolve` here.

lib/plugins/docker/index.js
```diff
@@ -4,13 +4,25 @@ import preset from '../../pattern-preset';
...
+  }
+
   getPattern() {
     return preset('Docker');
   }
 
   parseBlob(blob) {
-    insertLink(blob.el, DOCKER_FROM, {
-      resolver: 'dockerImage',
+    insertLink(this, blob.el, DOCKER_FROM, {
```
josephfrazier: Instead of passing the plugin as the first argument to `insertLink`, what if we just put it into the object literal, alongside `image`:

```js
insertLink(blob.el, DOCKER_FROM, {
  pluginName: this.constructor.name,
  image: '$1',
});
```

This way, `insertLink` won't have to pass an extra argument to `replace` and `buildDataAttr` at all. Note that doing this would invalidate my comment at https://github.com/OctoLinker/browser-extension/pull/274/files#r99657840, since the extra argument will no longer be present.

lib/plugins/python/index.js
```diff
@@ -1,16 +1,45 @@
...
+  }
+
   getPattern() {
     return preset('Python');
   }
 
   parseBlob(blob) {
-    insertLink(blob.el, PYTHON_IMPORT, {
-      resolver: 'pythonUniversal',
+    insertLink(this, blob.el, PYTHON_IMPORT, {
```
josephfrazier: (same comment as https://github.com/OctoLinker/browser-extension/pull/274/files#r99663118)

lib/plugins/docker/index.js
stefanbuck: The idea of passing the plugin to `insertLink` was to avoid duplication of `pluginName: this.constructor.name` across all plugins. To avoid conflicts with other prop key names passed to `insertLink` the name is prefixed with an underscore. 

Maybe it's not the best idea, since it hides the actual behaviour and makes it just more complicated. Thanks for pointing me in the right direction. That's why i believe and love code reviews 👍

stefanbuck:
Thanks for the great feedback. I'll follow up on your naming suggestions. 

Regarding the refactoring. Do you think it makes sense to open a refactor PR for every single plugin?  If we do that, we have to make it backwards compatible in order to keep the other plugins working. If you agree, I would like to open one big PR where every plugin refactoring is a commit. This should make it easier to review.

lib/plugins/docker/index.js
josephfrazier: Yeah, I was thinking about the duplication as well, when I was writing the comment. It's not ideal, but I think it's worth the gains in clarity that putting it in the map gives.

If you'd like, you could instead pass `plugin: this` to `insertLink`, and `insertLink` could delete the `plugin` key and add a `pluginName` key, to avoid duplicating `.constructor.name`, but I'm not sure if that's better...

I definitely agree on loving code reviews 👍

josephfrazier:
I think it'd be much better to have a single PR with one plugin per commit, as you've described. Making separate PRs doesn't seem like it's worth the effort of maintaining backwards-compatibility. I look forward to seeing the updates!

test/plugins/docker.test.js
```diff
@@ -1,38 +1,38 @@
 import assert from 'assert';
-import dockerImage from '../../lib/resolver/docker-image.js';
+import dockerImage from '../../lib/plugins/docker';
```
stefanbuck: For simplicity and keeping the PR small as possible, I just fixed the broken tests. I'll polish all plugin tests in a separate PR.

lib/plugins/bower-manifest.js
```diff
@@ -0,0 +1,61 @@
...
+
+    if (type === 'liveResolverQuery') {
+      return liveResolverQuery({ type: 'bower', target });
+    }
+
+    return [
+      gitUrl({ target }),
+      githubShorthand({ target }),
+    ];
+  }
```
stefanbuck: @josephfrazier what do you think about handling the different resolver types in one handler? It feels a bit inconvenient to me. In the old implementation this case was handled in a better way. 

One possible option would be to split the plugin by features. In this case one for linking the dependencies and another which is responsible for linking the `main` field. I'm not sure if this makes thing easier or more complicated.

stefanbuck:
I'm almost complete with this one. There is one outstanding question regarding the resolve handler function. Besides that I'm almost through. I'll do some minor improvements later this week.

lib/plugins/bower-manifest.js
josephfrazier: Are you talking about the calls to `JavaScriptPlugin.resolve`, `liveResolverQuery`, `gitUrl`, and `githubShorthand` being made explicitly, rather than implicitly like they are [here](https://github.com/OctoLinker/browser-extension/pull/274/files/e24c1ddec2445790d439dddfa09384ece592a520#diff-a8fe7166d47dcb744448dab0f2cf7befL11) and [here](https://github.com/OctoLinker/browser-extension/pull/274/files/e24c1ddec2445790d439dddfa09384ece592a520#diff-a8fe7166d47dcb744448dab0f2cf7befL21)? That doesn't really seem like a problem to me, but perhaps I'm misunderstanding the question. Splitting the Bower plugin into multiple other plugins sounds like it would just make things more complicated, especially if it's done as part of this refactor, rather than as a separate change.

stefanbuck: I'm talking about the if statements in the resolve function.  In my current implementation, the `type` attributes define which resolver is used. Without knowing how the click-handler works, I would assume that every `insertLink` can have his own click handler. What do you think? Maybe I'm over-thinking it.

stefanbuck: Let's ignore this for now. It's not really important

josephfrazier: Yeah, I think we can worry about that later if needed :)

stefanbuck:
@josephfrazier can you help with this? 

```
Downloading googlechrome 64 bit
  from 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'
Progress: 100% - Completed download of C:\Users\appveyor\AppData\Local\Temp\1\chocolatey\GoogleChrome\56.0.2924.87\googlechromestandaloneenterprise64.msi (48.26 MB).
Download of googlechromestandaloneenterprise64.msi (48.26 MB) completed.
Error - hashes do not match. Actual value was '1450DE87F0289CE725DF1C138BBA2666304FB94BF4476B03C06D058C999D4805'.
ERROR: Checksum for 'C:\Users\appveyor\AppData\Local\Temp\1\chocolatey\GoogleChrome\56.0.2924.87\googlechromestandaloneenterprise64.msi' did not meet 'c47ff551ff8b251268905cd87fb299959e6bcb3640a3e60085a3a7dbf176845f' for checksum type 'sha256'. Consider passing --ignore-checksums if necessary.
The install of googlechrome was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\GoogleChrome\tools\chocolateyInstall.ps1'.
 See log for details.
Chocolatey installed 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
Failures
 - googlechrome (exited -1) - Error while running 'C:\ProgramData\chocolatey\lib\GoogleChrome\tools\chocolateyInstall.ps1'.
 See log for details.
Did you know the proceeds of Pro (and some proceeds from other 
```

josephfrazier:
Hmm, that's a weird one. I just searched for the error and it looks like other people are seeing it too: https://chocolatey.org/packages/GoogleChrome#comment-3171878283

I don't really know why this would be happening, but it seems fine to add `--ignore-checksums` to the install command, considering it's being downloaded over HTTPS from Google.

stefanbuck:
@josephfrazier can I merge it by myself or is there anything left from your side?

josephfrazier:
This looks good to me. Excellent work on this! I'll let you do the honors of merging 👍

stefanbuck:
:smile: :tada:

Issue #278, opened 2017-02-06
Object.entries is not a function


xovel:
OctoLinker cannot work already.

> Object.entries is not a function

**Browser name:**
QQBrowser

**Browser version:** 
v9.5.3
Chromium 53.0.2785.104

**OctoLinker version:** 
4.8.0

**URL and line number where issue occurs:** 

storage.js:39Uncaught (in promise) TypeError: Object.entries is not a function(…)

**Expected behavior:** 

**Actual behavior:**

josephfrazier:
Hi @xovel, thanks for reporting the issue! I suspect this is related to the recent removal of `babel-polyfill` in https://github.com/OctoLinker/browser-extension/pull/258#issuecomment-275813737.

Are you able to update your browser to a more recent version? It looks like your current browser is a fork of an outdated version of Chromium, which may not support `Object.entries`. Furthermore, privacy and security issues have been discovered in QQ Browser, so I'd recommend switching to Chromium or Chrome if you can. See here for more info: https://citizenlab.org/2016/03/privacy-security-issues-qq-browser/

Whether or not you update your browser, it might be possible to fix this by reintroducing an `Object.entries` polyfill without causing errors about `Promise`. I'll leave this link here for reference: https://stackoverflow.com/questions/35090153/babel-support-for-object-entries

xovel:
Thanks for your reply in this issue @josephfrazier . The version of my QQBrowser is currently newest.

`Object.entries` is a new `ES7` features and `Chromium 53` does not support it. Unfortunally, I've get used to use this browser in my work, change it to another maybe result in a series of problems. Anyway, I also use `Chrome 55.0.2883.75` in my computer, and `OctoLinker 4.8.0` work very well in this browser.

I try to roll the `OctoLinker`'s version to `4.6.0` in QQBrowser, and it worked again.

PR #279, opened 2017-02-06
Use `core-js` to polyfill `Object.entries`

josephfrazier:
This fixes https://github.com/OctoLinker/browser-extension/issues/278

Note that `core-js` was already installed as a dependency of
`babel-polyfill`, so there are no npm-shrinkwrap.json changes. This
commit just adds back the `Object.entries` polyfill without interfering
with `Promise`, like having all of `babel-polyfill` would (see 050604b).

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -1,3 +1,4 @@
+import 'core-js/fn/object/entries';
 import OctoLinker from './octo-linker.js';
 import * as storage from './options/storage.js';
 
--- a/lib/click-handler.js
+++ b/lib/click-handler.js
@@ -1,3 +1,4 @@
+import 'core-js/fn/object/entries';
 import $ from 'jquery';
 import { showTooltip } from './gh-interface.js';
 import { flattenAndCompact } from './utils/array';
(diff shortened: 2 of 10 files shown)


josephfrazier:
Oops, that makes much more sense! I had originally used a non-polyfill version of `entries`:

```js
import entries from 'core-js/library/fn/object/entries';
```

but I forgot to remove duplicates after switching to the polyfill. It should be fixed as of https://github.com/OctoLinker/browser-extension/pull/279/commits/860958c5df1c6356283743f9b89208669d2ac606. Let me know if this looks good, and I'll clean up the commit message while merging.

stefanbuck:
Could this also replace `babel-polyfill` in our test setup? https://github.com/OctoLinker/browser-extension/blob/64199c1bc1de8bf9ddfd9bc35df38d0d37c1a9b4/test/main-lib.js#L1

josephfrazier:
Sure thing, I just pushed https://github.com/OctoLinker/browser-extension/pull/279/commits/0e939e1227e7c2587f59b99cbaab6056e1c209e7

I used 'core-js' instead of 'core-js/library/fn/object/entries' because PhantomJS is missing some other things like Map and Symbol, and I didn't want to track down all of them and import them individually, since bundle size isn't an issue for testing.

PR #281, opened 2017-02-13
Update `REQUIRE` regex to allow whitespaces around quotes

### LINKED ISSUE #236: Bug in require regex
It seems it fails to detect when there is spaces between parens and the string literal

something like

```js
var rollup = require( 'rollup' )
```

Not sure.

~~https://github.com/jonschlinkert/requires-regex may help~~ actually it not helps for that case

Arkni:
Fixes #236

### DIFF
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -1,6 +1,6 @@
 import go from './go.js';
 
-const REQUIRE = /require(?:\.resolve)?(?:\s|\()['"]([^'"\s]+)['"]\)?/g;
+const REQUIRE = /require(?:\.resolve)?(?:\s|\()\s*['"]([^'"\s]+)['"]\s*\)?/g;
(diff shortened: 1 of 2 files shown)


packages/helper-grammar-regex-collection/index.js
```diff
@@ -1,6 +1,6 @@
 import go from './go.js';
 
-const REQUIRE = /require(?:\.resolve)?(?:\s|\()['"]([^'"\s]+)['"]\)?/g;
+const REQUIRE = /require(?:\.resolve)?(?:\s|\()(?: +)?['"]([^'"\s]+)['"](?: +)?\)?/g;
```
josephfrazier: I think this should be changed to

```js
const REQUIRE = /require(?:\.resolve)?(?:\s|\()\s*['"]([^'"\s]+)['"]\s*\)?/g;
```

Using `\s*` instead of `(?: +)?` is shorter and it will match other whitespace characters like tabs. We should also add some tests with tab characters.

Arkni:
Done!

josephfrazier:
Great, thanks again!

Arkni:
You're welcome :)

PR #282, opened 2017-02-15
WIP: make helper-grammar-regex-collection more readable

josephfrazier:
This uses [XRegExp] to allow the regular expressions in
`helper-grammar-regex-collection` to:

1. be written in a [whitespace-agnostic] format
2. to share common [subpatterns] like `['"]([^'"\s]+)['"]`
3. to only capture [named groups], to avoid `(?:` everywhere

[XRegExp]: http://xregexp.com/
[whitespace-agnostic]: http://xregexp.com/flags/#extended
[subpatterns]: http://xregexp.com/api/#build
[named groups]: http://xregexp.com/flags/#explicitCapture

It takes advantage of [template literals] to allow the patterns to be
written as multi-line strings, but without having to double-escape
special characters.

[template literals]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

@stefanbuck, what do you think? In my opinion, this makes the regex
syntax a bit more verbose, but much easier to read and modify.

### DIFF
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -18,7 +18,12 @@ const regex = (pattern) => {
   return XRegExp.build(buildPattern, subpatterns, flags);
 };
 
-subpatterns.captureQuotedDep = regex`['"](?<$1>[^'"\s]+)['"]`;
+subpatterns.captureQuotedDep = regex`
+  ['"]            # beginning quote
+  (?<$1>[^'"\s]+) # capture the word inside the quotes
+  ['"]            # end quote
+`;
+
 subpatterns.importMembers = regex`[\r\n\s\w{},*\$]*`;
 subpatterns.from = regex`( \s from \s )`;
 
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -25,7 +25,7 @@ subpatterns.captureQuotedDep = regex`
 `;
 
 subpatterns.importMembers = regex`[\r\n\s\w{},*\$]*`;
-subpatterns.from = regex`( \s from \s )`;
+subpatterns.from = regex`\s from \s`;
 
 const REQUIRE = regex`
   require(\.resolve)?
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -31,7 +31,6 @@ const REQUIRE = regex`
   require(\.resolve)?
   ( \s | \( ) \s*
   {{captureQuotedWord}}
-  \s* \)?
 `;
 
 const IMPORT = regex`
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -1,5 +1,5 @@
-import XRegExp from 'xregexp';
 import go from './go.js';
+import regexBuilder from './regexBuilder';
 
 // The regular expressions in this file are built using XRegExp (http://xregexp.com/)
 //
@@ -12,11 +12,7 @@ import go from './go.js';
 
 const subpatterns = {};
 
-const regex = (pattern) => {
-  const buildPattern = `(?x)${String.raw(pattern)}`;
-  const flags = 'ngm';
-  return XRegExp.build(buildPattern, subpatterns, flags);
-};
+const regex = regexBuilder(subpatterns);
 
 subpatterns.captureQuotedWord = regex`
   ['"]            # beginning quote
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -1,17 +1,7 @@
 import go from './go.js';
 import regexBuilder from './regexBuilder';
 
-// The regular expressions in this file are built using XRegExp (http://xregexp.com/)
-//
-// Note that they are spread across multiple lines and that whitespace is irrelevant,
-// so spaces have to be matched with \s.
-// http://xregexp.com/flags/#extended
-//
-// Also, groups are not captured by default, so only named groups work, like (?<name>pattern)
-// http://xregexp.com/flags/#explicitCapture
-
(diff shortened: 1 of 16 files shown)


stefanbuck:
This looks awesome, well done! I'll take a closer look later

packages/helper-grammar-regex-collection/index.js
```diff
@@ -1,15 +1,81 @@
...
-const HOMEBREW = /(?:depends_on|conflicts_with)(?: cask:| formula:)? ['"]([^'"\s]+)['"]/g;
-const TYPESCRIPT_REFERENCE = /\/{3}\s?<reference path=['"]([^'"\s]+)['"]/g;
-const DOCKER_FROM = /FROM\s([^\n]*)/g;
-const VIM_PLUGIN = /(?:(?:(?:Neo)?Bundle(?:Lazy|Fetch)?)|Plug(?:in)?)\s['"]([^'"\s]+)['"]/g;
-const RUST_CRATE = /(?:extern crate|use) ([^:; ]+)/g;
-const PYTHON_IMPORT = /^\s*(?:import|from)\s([^\s]*)/gm;
+const subpatterns = {};
+const regex = regexBuilder(subpatterns);
+
+subpatterns.captureQuotedWord = regex`
```
stefanbuck: It feels a bit odd to me to modify `subpatterns` after passing it to `regexBuilder`. Just my personal preference.

packages/helper-grammar-regex-collection/index.js
```diff
@@ -1,15 +1,81 @@
...
-const EXPORT = /export [\r\n\s\w{},*\$]*(?: from )['"]([^'"\s]+)['"]/g;
-const GEM = /gem ['"]([^'"\s]+)['"]/g;
-const HOMEBREW = /(?:depends_on|conflicts_with)(?: cask:| formula:)? ['"]([^'"\s]+)['"]/g;
-const TYPESCRIPT_REFERENCE = /\/{3}\s?<reference path=['"]([^'"\s]+)['"]/g;
-const DOCKER_FROM = /FROM\s([^\n]*)/g;
-const VIM_PLUGIN = /(?:(?:(?:Neo)?Bundle(?:Lazy|Fetch)?)|Plug(?:in)?)\s['"]([^'"\s]+)['"]/g;
-const RUST_CRATE = /(?:extern crate|use) ([^:; ]+)/g;
-const PYTHON_IMPORT = /^\s*(?:import|from)\s([^\s]*)/gm;
+const subpatterns = {};
+const regex = regexBuilder(subpatterns);
```
stefanbuck: The result from `regexBuilder` is actually the `regexBuilder`, isn't it? Therefore I would prefer to rename `regex` to `regexBuilder` and `regexBuilder` to some different.

josephfrazier: Yeah, the naming is a little strange here. My perspective on it is that `regex` is a function that takes a string and returns a RegExp, much like its constructor: `RegExp('my pattern')`, so the `regex` name makes sense to me.

However, `regexBuilder` is still a little awkward. While it does build the `regex` function, it still sounds like it might do the same thing as `regex`. Besides, I discovered there's already another `regexBuilder` in the codebase ([lib/plugins/helper/regex-builder.js](https://github.com/OctoLinker/browser-extension/blob/cc1bae8b95d60d14c058ce1e610e84619395bfeb/lib/plugins/helper/regex-builder.js)). I not sure what a better name would be, though. Some ideas:

* `regexTagBuilder`
* `regexTagFunctionBuilder`

packages/helper-grammar-regex-collection/index.js
josephfrazier: I just pushed up a [refactor](https://github.com/OctoLinker/browser-extension/pull/282/files/54d73ecb030eb4a3a3f29a5a4ccf331f4ecd0caa..2c2d663e56642b35ec3d735d2ce7e241bbfcddc1) I had been working on, that addresses this by allowing normal template literal interpolation to be used, so that `subpatterns` isn't exposed outside of the `regex` tag function.

For what it's worth, this refactor was also done with the idea that I might eventually contribute it back to XRegExp. See https://github.com/slevithan/xregexp/issues/103

stefanbuck: Nice refactoring, looks damn good 👍 It's great to see your contribution to other project as well. Hopefully they will respond soon.

packages/helper-grammar-regex-collection/index.js
josephfrazier: What if we just get rid of of the `regexBuilder` function, and have `regex` uses the same flags every time (since the code calling it doesn't actually need to use different flags)? I just pushed up a commit showing what it could look like: https://github.com/OctoLinker/browser-extension/pull/282/commits/a3a25adad8d6b843227863799a0697f0e0c96334?w=true

packages/helper-grammar-regex-collection/index.js
josephfrazier: Thanks! Once we sort out the [naming question](https://github.com/OctoLinker/browser-extension/pull/282/files/54d73ecb030eb4a3a3f29a5a4ccf331f4ecd0caa#r101380958), I'll merge it.

If XRegExp later decides to incorporate the tag function, we'll be able to use it directly instead. Otherwise, I could always publish a separate module to make the tag more generally available.

packages/helper-grammar-regex-collection/index.js
stefanbuck: Look good and your explanation behind naming it `regex` makes sense to me as well.

josephfrazier: Thanks, I just merged in the changes! I'm still working on learning why [this](https://github.com/OctoLinker/browser-extension/blob/204d3d208a248c87be527b75e465449f4e04a8e6/packages/helper-grammar-regex-collection/regex.js#L19-L21) is necessary, but I wanted to go ahead and land the rest of it, since it would conflict with future changes to the regular expressions.

PR #283, opened 2017-02-15
Automatically fix lint errors

josephfrazier:
This uses eslint's [--fix] flag to repair minor code style issues on the
fly. Note that this flag is very [conservative], so it shouldn't cause any
problems.

[--fix]: http://eslint.org/docs/user-guide/command-line-interface#fix
[conservative]: https://github.com/eslint/eslint/issues/7873#issuecomment-271701235

EDIT: Actually, I just realized that since the `lint` script runs as part of `npm test`, this would cause Travis to allow style errors. I'll think about a better way to have autofixing locally, but not on CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   },
   "main": "lib/app.js",
   "scripts": {
-    "lint": "eslint .",
+    "lint": "eslint . --fix",
     "pretest": "npm run lint",
     "test": "karma start",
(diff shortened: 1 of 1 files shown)


stefanbuck:
I tend to delegate this to the developer in order to keep the process transparent for them.

josephfrazier:
That's fair :) There doesn't seem to be a clean way to avoid `--fix` on CI, anyway.

PR #285, opened 2017-02-17
Move Babel settings from webpack.config.js to package.json

josephfrazier:
This makes it possible for other tools to use the Babel settings. For
example, `babel-node` can be used during development as a REPL with
files in the repo:

    $ babel-node -p "require('./packages/helper-grammar-regex-collection/regex.js')"
    { default: [Function] }


stefanbuck:
Make sense

PR #286, opened 2017-02-17
Test on Chrome/Firefox instead of PhantomJS, if possible

josephfrazier:
This uses [karma-detect-browsers] to determine if Chrome and Firefox are
present locally, and to test against them if so. If it can't find either
browser, it will use PhantomJS as before.

Additionally, both Travis and AppVeyor will have Chrome and Firefox
installed, so CI tests will use browsers as well. See the following
links for more information about these changes:

* https://ariya.io/2015/09/javascript-testing-with-latest-firefox-and-chrome-on-appveyor
* https://swizec.com/blog/how-to-run-javascript-tests-in-chrome-on-travis/swizec/6647
* https://docs.travis-ci.com/user/firefox/

[karma-detect-browsers]: https://www.npmjs.com/package/karma-detect-browsers

---

Note that I had to slightly tweak a test to get it to pass on both
Chrome and Firefox, but it's a test that has been flaky on PhantomJS in
the past, so I don't think it's an issue.

I also disabled AppVeyor's node_modules cache, because `npm install` had
started to fail with:

    npm WARN retry will retry, error on last attempt: Error: EPERM: operation not permitted, rename 'C:\Users\appveyor\AppData\Local\Temp\1\npm-2400-004727be\registry.npmjs.org\glob\-\glob-7.1.1.tgz.3784351047' -> 'C:\Users\appveyor\AppData\Local\Temp\1\npm-2400-004727be\registry.npmjs.org\glob\-\glob-7.1.1.tgz'


.appveyor.yml
```diff
@@ -2,13 +2,15 @@
 environment:
   nodejs_version: "6"
 
-cache:
-  - node_modules
+#cache:
+# - node_modules
```
stefanbuck: Is this just a temporary fix?

josephfrazier: Yeah, it looks like AppVeyor was just being flaky, rather than the error having anything to do with this PR. I've just uncommented that bit (https://github.com/OctoLinker/browser-extension/pull/286/commits/98010f6e2c7662709e9db3b5f5d37ff7ccf2af59), and it still passes, so I'll go ahead and merge.

PR #295, opened 2017-02-27
Replace babel-preset-es2015 with babel-preset-env

stefanbuck:
The babel environment preset is a smarter preset that will add or remove babel transformation plugins based on your target environment.  Check out http://www.2ality.com/2017/02/babel-preset-env.html for details


package.json
```diff
@@ -33,7 +33,18 @@
...
     "presets": [
-      "es2015"
+      [
+        "env",
+        {
+          "targets": {
+            "browsers": [
+              "last 3 Chrome versions",
+              "last 3 Firefox versions",
+              "last 3 Opera versions"
```
stefanbuck: The last three versions should be enough I would say.

stefanbuck: Ideally we could specify the browser according to the current build target, but maybe a bit over engineered for our case

stefanbuck:
Damn .. I didn't think about the CI case

josephfrazier:
I've read about `babel-preset-env` before and it sounds great, so I'd love to help make this happen.

I think the AppVeyor failure is unrelated. The [error message](https://ci.appveyor.com/project/stefanbuck/browser-extension/build/1.0.274#L362) looks like the one I saw [here](https://github.com/OctoLinker/browser-extension/pull/286#issue-208578119).

As for Travis, it looks like [the Chromium version is pretty old](https://travis-ci.org/OctoLinker/browser-extension/builds/205902621#L690). I'm looking into how to upgrade it.

josephfrazier:
Ok, I added a commit which fixes the Travis build. I also tried to fix the AppVeyor build by disabling the node_modules cache, but that didn't help (here's the [build](https://ci.appveyor.com/project/josephfrazier/browser-extension/build/1.0.135)), so I didn't add it to this PR. It seems like the cache is unrelated, and AppVeyor is just having issues [again](https://github.com/OctoLinker/browser-extension/pull/286#discussion_r101907484).

npm-shrinkwrap.json
```diff
@@ -1858,6 +2232,820 @@
       "resolved": "http://registry.npmjs.org/chokidar/-/chokidar-1.6.1.tgz",
       "dev": true,
       "dependencies": {
+        "fsevents": {
```
josephfrazier: Oops, this sneaked back in. You can remove it by running

    rm -rf node_modules/chokidar/node_modules/fsevents
    npm shrinkwrap

You can also avoid it by making sure you add `--no-optional` to your `npm install` commands. It's a pain, but I'm not sure if there's a better way around it (unless you'd to try switching the project to another NPM client like [yarn](https://github.com/yarnpkg/yarn)?)

package.json
josephfrazier: 3 versions sounds good to me. If someone complains, we can always change it later.

Yeah, I think specifying only the current build target is overkill. If the goal is to reduce bundle size, there's easier things to do like minifying the dist/ .js files and/or removing sourcemaps from the out/ .zip files.

npm-shrinkwrap.json
stefanbuck: I've forgot to run `npm install` with `--no-optional` 😊 I've to admit that I never used yarn before. I'm happy to give it a try, finally.

josephfrazier: (I also forget `--no-optional` all the time...)

I've experimented with yarn on [a new project](https://github.com/josephfrazier/octopermalinker) that I forked from OctoLinker, and [it seems to work out ok](https://travis-ci.org/josephfrazier/octopermalinker/builds/205676836). Maybe I'll open a pull request, so we can talk about it there.

PR #296, opened 2017-02-28
Don't minify JS in release build

josephfrazier:
https://github.com/OctoLinker/browser-extension/pull/295 accidentally
broke `npm run package-all` because webpack runs uglifyjs, which doesn't
support arrow functions, etc (see https://github.com/mishoo/UglifyJS2/issues/448)

This change makes it so that webpack doesn't run uglifyjs, so that the
release build works again. This actually reduces the size of the .zip
files in out/, because the source maps are currently included.

Also, this makes Travis run `npm run package-all` on each build, like
AppVeyor does, to ensure that we know if release builds fail, even if
AppVeyor doesn't run them (this happened in
https://ci.appveyor.com/project/stefanbuck/browser-extension/build/1.0.278)


stefanbuck:
Good catch! It's funny that our build is smaller without minifying it.

stefanbuck:
Travis is facing some issues with AWS https://www.traviscistatus.com/

josephfrazier:
On build size: I was surprised by that too. I would have expected them to be equal-ish, but I guess that's not how it works.

On Travis: Yeah... so is https://githublinker.herokuapp.com/, for that matter :(

PR #297, opened 2017-02-28
Use yarn to install nodejs dependecies

stefanbuck:
As suggested by @josephfrazier, this PR replaces npm with yarn. Let's see how this goes


josephfrazier:
I'm curious to see how this turns out! A couple of thoughts:

* We should probably change [the appveyor install command](https://github.com/OctoLinker/browser-extension/blob/298eb679ade7daaa29f83b6a15030ce585bbbf9b/.appveyor.yml#L17-L18) as well. I [had to use](https://github.com/josephfrazier/octopermalinker/commit/1d6c61d6a0d9b06fcec60c724805534a852bb37a) `yarn install --ignore-engines` to get it working.
* The Contributing guide should be updated to use `yarn install` or just `yarn` instead of `npm install`. We might also want to point to https://yarnpkg.com for context.

stefanbuck:
According to the yarn docs, yarn is already preinstalled on AppVeyor https://yarnpkg.com/lang/en/docs/install-ci/#appveyor-tab

Good shout with the contributing guides! Will update them now

josephfrazier:
Yarn is preinstalled on AppVeyor, but it doesn't sound like AppVeyor will automatically run it like Travis does.

josephfrazier:
Oh, it looks like `yarn install` didn't need `--ignore-dependencies` after all. The build still failed, but that should be fixed by https://github.com/OctoLinker/browser-extension/pull/296

stefanbuck:
Unbelievable everything is green ... and blue 😉

josephfrazier:
Excellent! Now we'll just have to remember to use yarn instead of npm for changing dependencies.

Issue #300, opened 2017-03-04
Paths should fallback from `dist` to `lib` to `src`


josephfrazier:
**Browser name:** Chrome

**Browser version:** Version 56.0.2924.87 (64-bit)

**OctoLinker version:** d685d80a46238ba695d85210c8e451fbe9e1a5be

**URL and line number where issue occurs:** https://github.com/rauschma/enumify/blob/692021e4317b59b15e49090d1659e5c102004678/package.json#L4

**Expected behavior:** clicking on `./lib/enumify.js` (which doesn't exist) should navigate to [`./src/enumify.js`](https://github.com/rauschma/enumify/blob/692021e4317b59b15e49090d1659e5c102004678/src/enumify.js)

**Actual behavior:** unable to resolve the link

---

A common pattern in javascript projects is to store source code in either `src/` or `lib/`, then publish compiled code in either `lib/` or `dist/`. OctoLinker should try to assume this pattern is being used if it can't resolve a link to a file.

josephfrazier:
I ran across an edge case that I might try to fix later, so I'll reopen this. The problem here is that only the target path uses the fallback behavior, rather than the current file path joined with the target path.

https://github.com/jsdom/whatwg-url/blob/45ee04d40049c7a02ae3ad45e119dc49de14dc48/lib/public-api.js#L11

PR #303, opened 2017-03-04
javascriptFile: fallback from ./dist to ./lib to ./src

josephfrazier:
Fixes https://github.com/OctoLinker/browser-extension/issues/300


josephfrazier:
Haha, if that happens, we might be able to use their API to determine which files are present, instead of just trying them all.

PR #304, opened 2017-03-05
Add css import support

### LINKED ISSUE #290: [feature request] css @import linking
**Browser name:** Chrome

**Browser version:** Version 56.0.2924.87 (64-bit)

**OctoLinker version:** 4.8.0

**URL and line number where issue occurs:** N/A

**Expected behavior:** Link css `@import`'s

**Actual behavior:** OctoLinker does not interact with or anable linking for [`@import`](https://developer.mozilla.org/en-US/docs/Web/CSS/@import)'s. See [this css file](https://github.com/FormidableLabs/formidable-landers/blob/dccebf001535d77f469634a15ec29985ce73eff4/src/styles/styles.css) as an example. You should be able to click `"./base/variables.css"` within `@import "./base/variables.css";` and see that file.

stefanbuck:
Fix #290

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,11 +12,7 @@ To build and run the extension follow these steps.
 0. Clone the repository.
 0. Run `yarn install` to setup the project and install all required dependencies.
 0. Start hacking. If you're adding a plugin, you may be interested in seeing the commits that add these plugins:
-   * [Docker]
-   * [vimrc]
-   * [Rust]
-   * [Python]
-   * [project.json]
+   * [CSS]
 0. Build and load the extension:
    * Firefox (Quickstart):
      0. `yarn firefox-open`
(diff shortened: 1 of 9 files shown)


josephfrazier:
This looks great! I do have one nitpick, though. Would you mind moving the [fixture changes](https://github.com/OctoLinker/browser-extension/pull/304/files#diff-6629465044734f8eddee7564e186e06c) into a separate commit? Then, we can use the "Rebase and merge" button on Github and keep the commit history a little cleaner. Now that the regex syntax has been [refactored](https://github.com/OctoLinker/browser-extension/pull/282), the commit adding CSS import would be ideal to reference in [CONTRIBUTING.md](https://github.com/OctoLinker/browser-extension/blob/e7beabf26cf724c33b7290925c8a66defa5b4b33/CONTRIBUTING.md#quick-start), which is currently a bit outdated.

stefanbuck:
@josephfrazier I updated the contributing doc and remove some outdated parts. Later, we should add a more detailed version of the plugin architecture.

josephfrazier:
Sweet, thanks for updating the docs too! I just pushed a merge commit and I'm hoping that github's rebase/merge feature will just fast-forward to it, so it won't break the commit hash reference. Once CI passes, I'll try it.

More detailed architecture docs would be nice. I wonder if anyone's avoided getting involved due to lack of documentation...?

PR #305, opened 2017-03-05
Add option to open tabs in background

josephfrazier:
This lets the user specify the `active` argument to
[chrome.tabs.create]. If they uncheck the option, new tabs will open in
the background, rather than becoming focused. However, it defaults to
checked, which preserves the current behavior of focusing new tabs.

[chrome.tabs.create]: https://developer.chrome.com/extensions/tabs#method-create

### DIFF
--- a/lib/background/newTab.js
+++ b/lib/background/newTab.js
@@ -1,7 +1,8 @@
-function newTab(url, index) {
+function newTab(url, index, active) {
   chrome.tabs.create({
     url,
     index,
+    active,
   });
 }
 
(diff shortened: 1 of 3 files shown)


stefanbuck:
If I learned one thing from this project, it's that every user expects a different link behaviour after clicking a link. I'm happy to give them one more option 😜

josephfrazier:
Haha, thanks! I was browsing around and found a package that sounded interesting, and I wanted to look at it later rather than at the current time, so I decided to add this option. Ideally, it would be something that the user could choose on a per-link basis, but I don't know if there's an intuitive way of doing that, other than [using traditional links with hrefs](https://github.com/OctoLinker/browser-extension/pull/87) and letting the browser handle the UX (for example, Chrome and Firefox on OSX use command-click to open a background tab).

PR #258, opened 2017-01-25
Fix Unhandled promise rejection in storage.js

josephfrazier:
(this is built upon https://github.com/OctoLinker/browser-extension/pull/257 to make development easier)

See https://github.com/OctoLinker/browser-extension/issues/248

I decided to use [chrome-promise] instead of [chrome-storage-wrapper] or [chrome-storage-promise], because it is more popular and better maintained.

The commit messages have more detail, but to summarize: this doesn't quite work yet, seemingly due to some strange errors potentially related to the way the `chrome.storage` functions get wrapped into returning Promises (the same thing happens with the `chrome-storage-*` packages, for what it's worth):

<details>

```
16:09:28.308 Unhandled promise rejection TypeError: promise.then is not a function
Stack trace:
wrapPromise/<@resource://gre/modules/ExtensionCommon.jsm:328:9
Promise@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:4260:8
wrapPromise@resource://gre/modules/ExtensionCommon.jsm:327:14
callAsyncFunction@resource://gre/modules/ExtensionCommon.jsm:505:12
inject/</stub@resource://gre/modules/Schemas.jsm:1696:18
load/<@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:18747:12
notify/</run@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:4156:23
notify/<@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:4169:29
module.exports/flush@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:4524:10
MutationCallback*module.exports@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:4543:6
@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:4082:27
__webpack_require__@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:20:12
@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:213:2
__webpack_require__@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:20:12
@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:74:2
@file:///Users/josephfrazier/workspace/octolinker/browser-extension/dist/app.js:72:30
[...]

PR #306, opened 2017-03-06
Use async/await, cleanup fetch() code

josephfrazier:
This started out as an effort to use async/await instead of then/catch/callbacks, and as I went along, I realized that the `fetch` function could be called from the content script rather than the background script (see [MDN](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Content_scripts#Cross-domain_privileges)) and we could get rid of all the message passing wrappers around `fetch`, so I did that too. I also made `fetch` iterative rather than recursive.

Interestingly, even though using async/await requires babel to insert extra runtime code, the total extension size is smaller (probably because the background script no longer includes jquery):

    du -shc dist/*js

I tried to make each commit sensible on its own (some might be easier to view while ignoring whitespace), and to have the overall commit history follow my thought processes. Let me know if you have any questions/concerns!


stefanbuck:
I added the background script while I was working on adding Firefox support #100. At this time it was the only way to do requests. Great to see they getting closer to chrome extension. I need a bit more time to review this.

josephfrazier:
Oh, I see now why the background requests were necessary. Also glad that it has changed since then.

Take your time reviewing, I know there are some big changes here (and I was a little low on sleep when making them). I'll add some inline comments about parts that might need extra care.

lib/click-handler.js
```diff
@@ -110,19 +90,38 @@ function onClick(event) {
...
-      return console.error(err); // eslint-disable-line no-console
-    }
+  try {
+    // TODO remove this and fix tests
+    // Maybe use the following for mocking fetch?
+    // https://github.com/speedskater/babel-plugin-rewire
+    chrome.runtime.sendMessage({
+      type: 'fetch',
+      payload: urls,
+    });
```
josephfrazier: As mentioned in the comment, the tests currently stub out `sendMessage` and rely on it to be called. I briefly tried using https://github.com/speedskater/babel-plugin-rewire to do the same thing for our `fetch` module, but I wasn't able to get it working, so I just left things as they are for now.

test/click-handler.test.js
```diff
@@ -10,6 +10,7 @@ describe('click-handler', () => {
   let resolveFake;
 
   sandbox.stub(window.chrome.runtime, 'sendMessage');
+  window.console.error = () => {}; // Suppress fake onClick errors
```
josephfrazier: Before, the `return console.error(err);` line in `onClick` wasn't getting run because `sendMessage` was stubbed, so `tryLoad` never called its callback. Now, `console.error` gets called because fetch is an async function, so we just need to silence it here (to avoid having the tests print useless output). See my other comment for details about the stubbing: https://github.com/OctoLinker/browser-extension/pull/306/files#r104506996).

package.json
```diff
@@ -28,6 +28,12 @@
   },
   "babel": {
     "plugins": [
+      [
+        "transform-runtime",
+        {
+          "polyfill": false
```
josephfrazier: `babel-plugin-transform-runtime` is used instead of `babel-polyfill` due to
https://github.com/OctoLinker/browser-extension/pull/258#issuecomment-275813737

Note one of them is required. `babel-preset-env` alone isn't enough:
https://github.com/babel/babel/issues/5085#issuecomment-271640887

lib/click-handler.js
stefanbuck: In this case I would remove the code from here (since it isn't used anymore) and disable the related test until we find a suitable solution

lib/options/storage.js
```diff
@@ -12,43 +12,44 @@ options.forEach((item) => {
...
   };
 
-  // Firefox 51.0.1 doesn't have storage.sync, so wrap the property access in a
-  // Promise so that the exception can be handled.
-  return new Promise(resolve => resolve(chromep.storage.sync.set(data)))
-    .catch(() => chromep.storage.local.set(data));
+  try {
+    return await chromep.storage.sync.set(data);
+  } catch (err) {
+    return chromep.storage.local.set(data);
```
josephfrazier: Yes. I had originally put `await` in there, but eslint complained:

    /Users/josephfrazier/workspace/octolinker/browser-extension/lib/options/storage.js
      23:12  error  Redundant use of `await` on a return value  no-return-await

This makes sense because `async` functions always return a promise, so you don't have to `await` the return value. The reason that `await` is used inside the `try` block is so that any errors with `chromep.storage.sync.set(data)` will be caught inside this function, rather than being exposed to the function caller.

lib/click-handler.js
josephfrazier: Agreed, fixed in https://github.com/OctoLinker/browser-extension/pull/306/commits/af77eda38aab5c0aaa186c07db2261cd49f2345e

lib/background/fetch.js
```diff
@@ -1,48 +1,23 @@
...
-    cb(null, url, res);
-  }).fail(() => {
-    if (urls.length === 0) {
-      chrome.runtime.sendMessage({
-        type: 'track',
-        payload: {
-          category: 'fetch',
-          action: 'error',
-        },
+export default async (urls) => {
```
josephfrazier: Good point, fixed in https://github.com/OctoLinker/browser-extension/pull/306/commits/b896d9dea3d43eb4cf405337c34ea15104e94feb

josephfrazier:
Thanks for the review! I addressed the comments and merged this in. I also created https://github.com/OctoLinker/browser-extension/issues/311 for fixing the click-handler/fetch tests.

PR #307, opened 2017-03-06
Add HTML import support

stefanbuck:
Adding HTML import was an easy thing to do. The tricky part is to agree on which file extension we would like to support.

@tunnckoCore was providing the following

- highlight-text-html-basic // => .html, .html
- highlight-text-html-twig
- highlight-text-html-slim
- highlight-text-html-vue
- highlight-text-html-js // => ejs, and probably others
- highlight-text-html-handlebars // => hbs, handlebars, htmlbars
- highlight-text-html-erb
- highlight-text-jade // => jade, pug

what else do we need to add?

[Demo page](https://github.com/Polymer/polymer-analyzer/blob/b438736929167351a212d11a023c1c8849cdcad9/src/test/static/diamond/a.html)

### DIFF
--- a/lib/pattern-preset.js
+++ b/lib/pattern-preset.js
@@ -92,6 +92,16 @@ export const presets = {
       'highlight-source-css',
     ],
   },
+
+  html: {
+    pathSubstrings: [
+      '.html',
+      '.htm',
+    ],
+    githubClasses: [
+      'type-html',
+    ],
+  },
 };
 
 export default function (presetName, siblingPresetName) {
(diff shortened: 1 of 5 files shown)


tunnckoCore:
I think they are pretty enough and handles most of the cases. I just researched only these, there may be more.

Where you see the `type-html` class? And why you added only `highlight-text-html-basic`? All they are "html" from which you should detect rel=imports

stefanbuck:
Most of the times the class `type-html` is not needed. It only kicks in, if the source you look at doesn't contain a file extensions.

Initial I thought we should add all `highlight-` classes that @tunnckoCore suggested. After thinking for a while, I tend to not add them. The `highlight-` classes are used in snippets only for example within comments or markdown files. In that case the linking won't work anyway, because the relative path can't be built. 

We have the same issue for file extensions like `.jade` or `.handlebars`. A template file can be located anywhere and the referenced import path doesn't have to be necessary relative correct to this file.

PR #308, opened 2017-03-09
Haskell support

ousennin:
I guess I have to write a 
> convincing description of your PR and why we should land it

I shall be honest - I'm far from being sure that you should land it. First of all, my solution is partial and based on certain conventions and assumptions: it doesn't scan *.cabal file as it should. It simply tries to get several most possible urls and it does this successfully. Next: as for linking to Haskell documentation  - I've hardcoded access to base-4.9.1.0 package documentation and that's it. Unfortunately, it seems to me that package naming convention is not well suited for creating docs url out if. I may be mistaken, of course.
If you think it's a good idea to accept this partial solution (e.g. as a temporary) - then go for it! If no - well, it was fun to try solving this problem :). Maybe you could point me out a direction in which I should put my efforts to make this solution really good?


packages/helper-grammar-regex-collection/index.js
```diff
@@ -74,6 +74,12 @@ const PYTHON_IMPORT = regex`
   (?<$1>[^\s]*)
 `;
 
+<<<<<<< HEAD
```
r-obert: looks like unresolved merge conflict.

lib/plugins/haskell.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default class Haskell {
+  static resolve({ path, target }) {
+    const [, user, repo] = path.split('/');
+    const filePath = target.replace(/\./g, '/');
+    return [
+      `{BASE_URL}/${user}/${repo}/blob/master/src/${filePath}.hs`,
+      `{BASE_URL}/${user}/${repo}/blob/master/lib/${filePath}.hs`,
+      `{BASE_URL}/${user}/${repo}/blob/master/${filePath}.hs`,
+      `https://hackage.haskell.org/package/base-4.9.1.0/docs/${target.replace(/\./g, '-')}.html`,
```
josephfrazier: I think this can be replaced by `https://hackage.haskell.org/package/base/docs/${target.replace(/\./g, '-')}.html`, which will redirect to the `base-4.9.1.0` url, but would presumably stay up-to-date. For example: https://hackage.haskell.org/package/base/docs/Data-List.html

lib/plugins/haskell.js
```diff
@@ -0,0 +1,28 @@
...
+import preset from '../pattern-preset';
+
+export default class Haskell {
+  static resolve({ path, target }) {
+    const [, user, repo] = path.split('/');
+    const filePath = target.replace(/\./g, '/');
+    return [
+      `{BASE_URL}/${user}/${repo}/blob/master/src/${filePath}.hs`,
+      `{BASE_URL}/${user}/${repo}/blob/master/lib/${filePath}.hs`,
+      `{BASE_URL}/${user}/${repo}/blob/master/${filePath}.hs`,
```
josephfrazier: Are these file path conventions pretty strong in the Haskell community? I'm mostly just curious (OctoLinker uses heuristics in [other places](https://github.com/OctoLinker/browser-extension/issues/300), so it shouldn't be a problem here).

lib/plugins/haskell.js
ousennin: A good point!

packages/helper-grammar-regex-collection/index.js
```diff
@@ -105,4 +110,5 @@ export {
   CSS_IMPORT,
   HTML_IMPORT,
   go,
+  HASKELL_IMPORT,
```
stefanbuck: Can you add some tests to verify that the regex is working as expected, please. It's easy and it shouldn't take much time [see](https://github.com/ousennin/browser-extension/blob/master/packages/helper-grammar-regex-collection/test.js#L168-L183)

lib/plugins/haskell.js
ousennin: Well, I've came to this conclusion after analyzing a number of haskell source code repos.

ousennin:
I've made required changes.

josephfrazier:
Thanks for the quick follow-up, @ousennin! I've merged your changes. Like @stefanbuck said, we really appreciate the contribution :)

PR #316, opened 2017-03-18
Fetch and display release highlight

stefanbuck:
I thought it would be helpful to display the release highlight in the notification, so I created this PR.

screen shot 2017-03-18 at 18 28 14

The first release notes line defines the highlight and will be displayed in the notification.

### DIFF
--- a/lib/gh-interface.js
+++ b/lib/gh-interface.js
@@ -8,24 +8,13 @@ export function showTooltip($target, msg) {
   $target.attr('aria-label', msg);
 }
 
-function getUserName() {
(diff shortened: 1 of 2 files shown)


josephfrazier:
We'll have to remember to enter the highlight line in the release notes, but this looks good to me. I also added a commit that simplifies the code using async/await. If that's ok with you, go ahead and merge.

stefanbuck:
I totally missed the async/await use case here.

Thanks @josephfrazier for the additional commit.

PR #317, opened 2017-03-19
Add partial support for sass

stefanbuck:
This PR adds support for relative sass imports. It doesn't take [includePaths](https://github.com/sass/node-sass#includepaths) into account. 

[Demo](https://github.com/twbs/bootstrap/blob/v4-dev/scss/bootstrap-grid.scss)

**TODO**
- [x] Add tests
- [x] Document feature

### DIFF
--- a/lib/plugins/sass.js
+++ b/lib/plugins/sass.js
@@ -1,0 +1,31 @@
+import { join } from 'path';
+import pathParse from 'path-parse';
+import { CSS_IMPORT } from '../../packages/helper-grammar-regex-collection/index.js';
+import insertLink from '../insert-link';
+import preset from '../pattern-preset';
+import relativeFile from '../resolver/relative-file.js';
+
+export default class Sass {
+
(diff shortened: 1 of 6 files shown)


lib/plugins/sass.js
```diff
@@ -0,0 +1,31 @@
...
+import relativeFile from '../resolver/relative-file.js';
+
+export default class Sass {
+
+  static resolve({ path, target }) {
+    const { dir, base, ext } = pathParse(target);
+    const prefixedTarget = join(dir, `_${base}`);
+
+    return [
+      relativeFile({ path, target: `${prefixedTarget}.scss` }),
```
josephfrazier: Should we also try the '.sass' extension here?

stefanbuck: Good point. I removed `${prefixedTarget}.${ext}` in favour of the additional file extension `.sass`

PR #321, opened 2017-03-21
houndci test

stefanbuck:
PR to verify that HoundCI is working as excepted.


josephfrazier:
Hmm, I expected Hound to have commented, since there are some style errors: https://travis-ci.org/OctoLinker/browser-extension/builds/213455957#L405-L419

stefanbuck:
I dropped an email to their support team. Will keep you updated.

stefanbuck:
Finally I can report the latest news on this. They investigate into that issue over the last days and it seems like some rule in our eslint config breaks houndci. They suggested to remove some rules to have it basicly aligned with there supported eslint config. This isn't really an option and therefore I'll remove the houndci integration soon.

josephfrazier:
Oh, that's interesting... I'm curious, which rules were incompatible with their service? I definitely agree we shouldn't compromise our linting just to be able to have Hound running, especially since Travis runs the lint script as well.

PR #324, opened 2017-03-22
helper-grammar-regex-collection: dedupe exports

josephfrazier:
This way, we don't have to have that big export statement at the
bottom of the file, with all of the variable names repeated.


stefanbuck:
Good idea :+1:

PR #325, opened 2017-03-22
Less: support `@import` keywords

josephfrazier:
This follows up on commit 8736eac (#322)

See http://lesscss.org/features/#import-options

Demo page: https://github.com/less/less-docs/blob/315c5b3633d2ff2b94054574cfdf57132aa95eed/styles/index.less#L13


stefanbuck:
Awesome!!! Thank you

PR #327, opened 2017-03-22
Fallback to fuzzy search

stefanbuck:
Today, I tried to implement support for [less includePaths](http://lesscss.org/usage/) and [sass includePaths](https://github.com/sass/node-sass#includepaths). As expect this is really tricky to implement and maybe even impossible. There is no single point of truth where this property is defined. It could be everywhere. Of course we could search  in common places like in `gruntfile.js` or `gulpfile.js` file, but even that isn't entirely accurate. If the path isn't stored a string we're screwed.

Luckily, an idea came to my head and thought giving it a try. Why not using GitHub`s [search API](https://developer.github.com/v3/search/). I know that this isn't very accurate either, but at least it works in many (hopefully in almost every) case. 

I need a bit more time to think about it. I opened the PR to share the basic idea with you. What do you think @josephfrazier

[This](https://github.com/less/less-docs/blob/master/styles/index.less#L12) import wasn't working before.

### DIFF
--- a/lib/plugins/haskell.js
+++ b/lib/plugins/haskell.js
@@ -1,6 +1,7 @@
 import { HASKELL_IMPORT } from '../../packages/helper-grammar-regex-collection/index.js';
 import insertLink from '../insert-link';
 import preset from '../pattern-preset';
+import githubSearch from '../resolver/github-search.js';
 
 export default class Haskell {
   static resolve({ path, target }) {
@@ -10,6 +11,7 @@ export default class Haskell {
       `{BASE_URL}/${user}/${repo}/blob/master/src/${filePath}.hs`,
       `{BASE_URL}/${user}/${repo}/blob/master/lib/${filePath}.hs`,
       `{BASE_URL}/${user}/${repo}/blob/master/${filePath}.hs`,
+      githubSearch({ path, target: `${filePath}.hs` }),
       `https://hackage.haskell.org/package/base/docs/${target.replace(/\./g, '-')}.html`,
     ];
   }
--- a/lib/click-handler.js
+++ b/lib/click-handler.js
@@ -52,6 +52,12 @@ function getResolverUrls(dataAttr) {
       return null;
     }
 
+    if (typeof url === 'function') {
+      return {
+        func: url,
+      };
+    }
+
     if (typeof url === 'object') {
       return {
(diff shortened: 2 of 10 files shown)


josephfrazier:
I really like this idea! As long as it remains a fallback for when the other resolution methods fail, it seems like a strict improvement over the current behavior. Of course, if we find a counter-example where the search actually resolves to the _wrong_ file (maybe the repo has multiple files with the same name), we might have to refine things a little. Until then though, I think this is great!

stefanbuck:
I renamed to resolver to `github-search`,  as this describes what it does. Besides that I tweaked the API call in order to improve the result. If a file extension is present in `target` this will be set as query parameter which will hopefully results in a more accurate behaviour. 

I also updated the sass, less and haskell plugin to make use of the new resolver.  Some tests are flaky and needs to be refactored soon. @josephfrazier if you don't mind, I would like to do this in a later PR.

lib/plugins/haskell.js
```diff
@@ -10,6 +11,7 @@ export default class Haskell {
       `{BASE_URL}/${user}/${repo}/blob/master/src/${filePath}.hs`,
       `{BASE_URL}/${user}/${repo}/blob/master/lib/${filePath}.hs`,
       `{BASE_URL}/${user}/${repo}/blob/master/${filePath}.hs`,
+      githubSearch({ path, target: `${filePath}.hs` }),
```
stefanbuck: @ousennin I updated the haskell plugin in order to improve the coverage. This new resolver will perform a github search on the current repo in order to find the linked file. 

I didn't removed the other path definitions, since they are still valid in many cases and more accurate.

stefanbuck:
I just noticed that the search API doesn't work for private repos without an `access_token`. I'll add a new field to the settings page so every user can easily generate and add an `access_token`.

lib/plugins/haskell.js
josephfrazier: This seems fine to me :)

josephfrazier:
This looks good to me so far! I agree on refactoring tests in a separate PR. Did you want to add the `access_token` setting here, or should we go ahead and merge, then do that separately as well?

stefanbuck:
Initially I was thinking of adding the `access_token` within this PR. A separate PR is easier to review so I'll do it separately.

Issue #326, opened 2017-03-22
Support for Ruby require_relative


MatthewDG:
Hi all, thank you for the amazing extension. I saw that support for Ruby `require` and Ruby `require_relative` was addressed in issue #43 and applied in 4.0.0.  `require` links work fine now but there is still no support for `require_relative`.

josephfrazier:
Since [the Ruby plugin](https://github.com/OctoLinker/browser-extension/blob/e25f90da95ea67b84dfb63967de64f46504a8a4b/lib/plugins/ruby.js#L2) uses the [`REQUIRE` regex](https://github.com/OctoLinker/browser-extension/blob/e25f90da95ea67b84dfb63967de64f46504a8a4b/packages/helper-grammar-regex-collection/index.js#L14), fixing this should be very similar to the process described in https://github.com/OctoLinker/browser-extension/issues/315#issuecomment-287835601. [This line](https://github.com/OctoLinker/browser-extension/blob/e25f90da95ea67b84dfb63967de64f46504a8a4b/packages/helper-grammar-regex-collection/index.js#L15) should be made to match `require_relative`, and the [valid](https://github.com/OctoLinker/browser-extension/blob/e25f90da95ea67b84dfb63967de64f46504a8a4b/packages/helper-grammar-regex-collection/test.js#L52-L70) and [invalid](https://github.com/OctoLinker/browser-extension/blob/e25f90da95ea67b84dfb63967de64f46504a8a4b/packages/helper-grammar-regex-collection/test.js#L102-L105) test cases should be copied to use `require_relative` instead of `require`.

MatthewDG:
Awesome, thank you. I can go ahead and get started on it if it's alright with you.

josephfrazier:
That'd be great! Feel free to open a pull request even if it isn't working yet, and I'd be happy to help with any issues that arise.

MatthewDG:
I just about have it.  Changing `( require(\.resolve)? | import )` to `( require(\.resolve)? | import | require_relative )` does change require_relative lines to links.  Since Ruby `require` and `require_relative` do not need you to add '.rb' at the end, I noticed that links that do not include the .rb execute the `liveResolverQuery` in the Ruby plugin with `type: 'rubygems'`. For example, `require_relative 'example'` and `require 'example'` would try to link me to a website or repo for a Ruby Gem although it should be a file link. The links work fine when a directory is indicated, though (i.e. `require_relative 'lib/example'`). 

[...]

PR #328, opened 2017-03-23
Ruby require_relative support

MatthewDG:
This follows up on issue #326. `require_relative` links are clickable with this PR but do not always work. For example, the first link on <a href="https://github.com/MatthewDG/ormellow/blob/master/lib/sql_object.rb">this page</a> tries to resolve a path to a page dedicated to a Ruby gem, but the other two require_relative links work fine. On many other pages such as <a href="https://github.com/rails/rails/blob/master/actioncable/test/subscription_adapter/redis_test.rb">this one</a>, the links just throw a 404 error in the console.


stefanbuck:
Thanks man, looks good. Will be part of the upcoming release next week.

PR #330, opened 2017-03-26
Add field for github api token to options page

stefanbuck:
Besides adding the capability to specify an api token to improve the quality for Sass, Less and Haskell, this PR also includes a couple of minor change around the options page. 

Let's start with the new layout. I added [primer css](http://primercss.io) (github's style guide) to polish the layout of the options page. Also I tweaked  the option descriptions to make them easier to understand. Finally I removed the debug options.

### New layout
screen shot 2017-03-26 at 19 42 09

### Before

screen shot 2017-03-26 at 19 41 52


assets/options.html
```diff
@@ -3,23 +3,31 @@
...
...hadow:none}.btn-purple .counter{color:#683cb8;background-color:#fff}.btn-blue{color:#fff;background-color:#0361cc;background-image:-webkit-linear-gradient(top, #0679fc 0%, #0361cc 90%);background-image:linear-gradient(-180deg, #0679fc 0%, #0361cc 90%)}.btn-blue:focus,.btn-blue.focus{box-shadow:0 0 0 0.2em rgba(6,121,252,0.3)}.btn-blue:hover,.btn-blue.hover{background-color:#035cc2;background-image:-webkit-linear-gradient(top, #0374f4 0%, #035cc2 90%);background-image:linear-gradient(-180deg, #0374f4 0%, #035cc2 90%);background-position:0 -0.5em;border-color:rgba(27,31,35,0.5)}.btn-blue:active,.btn-blue.selected{background-color:#045cc1;background-image:none;border-color:rgba(27,31,35,0.5);box-shadow:inset 0 0.15em 0.3em rgba(27,31,35,0.15)}.btn-blue:disabled,.btn-blue.disabled{color:rgba(255,255,255,0.75);background-color:#81b0e5;background-image:none;border-color:rgba(27,31,35,0.2);box-shadow:none}.btn-blue .counter{color:#0366d6;background-color:#fff}.btn-danger{color:#cb2431;background-color:#fafbfc;background-image:-webkit-linear-gradient(top, #fafbfc 0%, #eff3f6 90%);background-image:linear-gradient(-180deg, #fafbfc 0%, #eff3f6 90%)}.btn-danger:focus{box-shadow:0 0 0 0.2em rgb
```
josephfrazier: Instead of pasting all this stuff in, what if we just add a couple link elements to the https://unpkg.com versions of the relevant files in the published npm packages? We can use [subresource integrity](https://www.srihash.org/) to verify that they do not change. This would look like:

```html
<link rel="stylesheet" href="https://unpkg.com/primer-css@5.0.0/build/build.css" integrity="sha384-Bg/C6ZDEvQjQtboCzdJbsH6QCir2Z4k57aGBdganyzSMwOTKqQJUjU4y6IN/Su6G" crossorigin="anonymous">
<link rel="stylesheet" href="https://unpkg.com/primer-forms@1.0.6/build/build.css" integrity="sha384-V1f3Tj00bC3QZNxrIbWHpggMZpFzizEQoTVf7lDfJ1ktkuokeoy/yntf187gJXo4" crossorigin="anonymous">
```

stefanbuck: I have to check how we can use primer css without adding the overhead of sass.

lib/options/options.js
```diff
@@ -1,44 +1,44 @@
...
   {
     type: 'checkbox',
-    name: 'debugMode',
-    label: 'Debug mode',
-    value: undefined,
-    defaultValue: false,
+    name: 'doTrack',
+    label: 'Usage reports',
+    description: 'Send anonymous usage reports to Google Analytics.',
+    defaultValue: true,
```
josephfrazier: Was this option moved to the bottom intentionally? If not, can we put it back above the previous one, to keep the diff cleaner?

assets/options.html
josephfrazier: Alternatively, we could `yarn add primer-core primer-forms` and copy their `build/build.css` files into `/assets` during the build process.

lib/options/options.js
stefanbuck: No, it was on purpose. I think the tracking options is less important and a lot of other apps are  putting this option to the end as well.

assets/options.html
stefanbuck: Thanks for all this options. I'll check out which works best for us. I have to check out subresource integrity, I have never heard about it before.

lib/options/options.js
josephfrazier: Ah, that makes sense.

assets/options.html
stefanbuck: I went with the last solution. Mainly because in one of the previous reviews, mozilla complained about remote scripts in our fixtures files and I'm unsure there view on remote css files.

josephfrazier:
I made a tiny tweak to make the new field empty by default, but this looks good to me now 👍

Issue #331, opened 2017-03-27
Doesn't link requirements.txt entries


josephfrazier:
**Browser name:** Chrome

**Browser version:** Version 57.0.2987.98 (64-bit)

**OctoLinker version:** 7c785a581facd284519c70f2abad18b11855d8ce

**URL and line number where issue occurs:** https://github.com/bndr/pipreqs/blob/1a990b52d5e1d24cd5a2e7673077ff48ae52f9d3/requirements.txt

**Expected behavior:** `wheel`, `Yarg`, and `docopt` are links

**Actual behavior:** they aren't

---

This was mentioned in a [review](https://addons.mozilla.org/en-US/firefox/addon/octolinker/reviews/847841/) on addons.mozilla.org

> It's a really great extension. I'm only missing two features:
> - linking requirements.txt entries
> - and working on gitlab.com and private gitlab instances

PR #332, opened 2017-03-27
Add requirements.txt plugin

josephfrazier:
Specification: https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format

Test page: https://github.com/bndr/pipreqs/blob/1a990b52d5e1d24cd5a2e7673077ff48ae52f9d3/requirements.txt

Fixes https://github.com/OctoLinker/browser-extension/issues/331

### DIFF
--- a/lib/pattern-preset.js
+++ b/lib/pattern-preset.js
@@ -77,6 +77,11 @@ export const presets = {
     ],
   },
 
+  RequirementsTxt: {
+    pathSubstrings: ['requirements.txt'],
+    githubClasses: [],
+  },
+
   Go: {
     pathSubstrings: ['.go'],
     githubClasses: [
--- a/lib/plugins/requirements-txt.js
+++ b/lib/plugins/requirements-txt.js
@@ -1,0 +1,28 @@
+import { REQUIREMENTS_TXT } from '../../packages/helper-grammar-regex-collection/index.js';
+import insertLink from '../insert-link';
+import preset from '../pattern-preset';
+import liveResolverQuery from '../resolver/live-resolver-query.js';
+
+export default class RequirementsTxt {
+
+  static resolve({ target }) {
+    return [
+      liveResolverQuery({
+        target,
+        type: 'pypi',
+      }),
+    ];
+  }
+
+  getPattern() {
+    return preset('RequirementsTxt');
+  }
+
+  parseBlob(blob) {
+    insertLink(blob.el, REQUIREMENTS_TXT, {
+      pluginName: this.constructor.name,
+      target: '$1',
+      path: blob.path,
+    });
+  }
+}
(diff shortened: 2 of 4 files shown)


lib/plugins/requirements-txt.js
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  parseBlob(blob) {
+    insertLink(blob.el, REQUIREMENTS_TXT, {
+      pluginName: this.constructor.name,
+      target: '$1',
+      path: blob.path,
+    });
+  }
+}
```
stefanbuck: Did your considered to extend the existing python plugin with this functionality?

josephfrazier:
Note that this currently doesn't support filenames like [dev-requirements.txt](https://github.com/alanhamlett/pip-update-requirements/blob/c8d64c8c865be48f1654052a546f68f37e718ff9/dev-requirements.txt), due to the way filename-based detection works.

lib/pattern-preset.js
```diff
@@ -77,6 +77,11 @@ export const presets = {
     ],
   },
 
+  RequirementsTxt: {
+    pathSubstrings: ['requirements.txt'],
+    githubClasses: [],
```
stefanbuck: Is there any reason why you didn't specified the githubClass?

josephfrazier: Github uses the `type-text` class for requirements.txt files, which seems too general to use for filetype detection.

stefanbuck: I don't think this will be an issue. It will just invoke the plugin and if there is no match with the regex it will not harm the user. Does it make sense to you?

lib/plugins/requirements-txt.js
josephfrazier: Yes, but I think it makes more sense to keep it separate, since the resolution logic is different. In particular, we don't need to do any local file, init, or api detection (all the stuff [here](https://github.com/OctoLinker/browser-extension/blob/7c785a581facd284519c70f2abad18b11855d8ce/lib/plugins/python.js#L10-L26)).

Additionally, since the python resolver [prefers](https://github.com/OctoLinker/browser-extension/blob/7c785a581facd284519c70f2abad18b11855d8ce/lib/plugins/python.js#L29) Python API docs over userland packages, we could end up resolving a requirements.txt entry of (for example) `argparse` to https://docs.python.org/3/library/argparse.html instead of https://github.com/ThomasWaldmann/argparse/. This is the opposite of the problem described at https://github.com/OctoLinker/browser-extension/pull/210

lib/pattern-preset.js
josephfrazier: Ah, that could allow it to work with alternative filenames like dev-requirements.txt (see  https://github.com/OctoLinker/browser-extension/pull/332#issuecomment-289582089), but I'm concerned that there might be false positives in other `type-text` blobs completely unrelated to python.

lib/plugins/requirements-txt.js
stefanbuck: Alright, thanks for the explanation

Issue #235, opened 2016-12-22
Support `jsnext:main` and `module` in package.json


tunnckoCore:
I believe simple they can be added here [lib/plugins/npm-manifest/index.js#L34-L44](https://github.com/OctoLinker/browser-extension/blob/c2ceab4373a7d297bf2b52fc72dafde57276df66/lib/plugins/npm-manifest/index.js#L34-L44).

Or at least, one of them. I believe [`jsnext:main`](https://github.com/rollup/rollup/wiki/jsnext:main) was suggested by Rollup creator. But `module` would become "standard" one day - if I'm not in mistake it already can be seen `<script type="module">` in some places.

josephfrazier:
Thanks for the suggestion! As you've noticed, it should be pretty straightforward to implement. Would you like to try it? [CONTRIBUTING.md](https://github.com/OctoLinker/browser-extension/blob/master/CONTRIBUTING.md) has a development guide, and I'd be happy to help with any issues you encounter.

For reference, here are some previous pull requests that have changes very similar to the ones that are required here:

* https://github.com/OctoLinker/browser-extension/pull/126
* https://github.com/OctoLinker/browser-extension/pull/127
* https://github.com/OctoLinker/browser-extension/pull/204

PS: There shouldn't be any problem with linking both `jsnext:main` and `module` fields.

tunnckoCore:
thanks for the links

PR #335, opened 2017-03-28
package.json plugin: link `module` and `jsnext:main` fields

josephfrazier:
Test page: https://github.com/northbrookjs/northbrook/blob/84a31d4f54db97b7448216b9bb04fc4671e28092/src/plugins/exec/__test__/a/package.json#L5-L6

Fixes https://github.com/OctoLinker/browser-extension/issues/235


stefanbuck:
@josephfrazier is on fire 😄

PR #340, opened 2017-04-07
test: Dedupe grammar tests

josephfrazier:
This allows the `go` tests to be handled alongside the others, rather
than duplicating the test logic. The diff is best viewed ignoring
whitespace changes: https://github.com/OctoLinker/browser-extension/pull/340/files?w=true


stefanbuck:
:+1:

PR #341, opened 2017-04-07
Extract common parseBlob() from plugins

josephfrazier:
Since so many of the plugins have the same `parseBlob` implementation
(just with a different set of regexes), let's make it so that the
plugins can just specify the regexes to use with `insertLink`

### DIFF
--- a/lib/octo-linker.js
+++ b/lib/octo-linker.js
@@ -3,6 +3,7 @@ import notification from './notification';
 import BlobReader from '../packages/blob-reader';
 import clickHandler from './click-handler';
 import Plugins from './plugin-manager.js';
+import insertLink from './insert-link';
 import debugMode from './debug-mode.js';
 import loadPlugins from './load-plugins';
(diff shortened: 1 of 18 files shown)


josephfrazier:
Thanks for the review.

> Did you considered​ to create a `BasePlugin` class which every plugin extends?

I hadn't considered it, but it be a good way to abstract the default `parseBlob` behavior out of `lib/octo-linker.js`. On the other hand, I had actually been thinking about making the plugins just regular objects instead of classes, since we don't actually need separate instances of them anyway. Maybe I'll push up a separate pull request with that change, and we can talk about the pros/cons of each approach.

> Therefore I was thinking if `getLinkRegexes` should be able to handle a both a single string and an array of strings.

Good point, I've pushed https://github.com/OctoLinker/browser-extension/pull/341/commits/c79d199448fcbe29afac5d6f92964ab482301410 to make this possible.

PR #344, opened 2017-04-10
Make plugins regular objects instead of classes

josephfrazier:
As mentioned in https://github.com/OctoLinker/browser-extension/pull/341#issuecomment-293080486,
this lets us avoid unnecessarily instantiating the class, and just use
the plugin directly instead.


josephfrazier:
Thanks! Yeah, I think it's a little more intuitive, especially for people who don't have much experience with ES6 classes.

PR #347, opened 2017-04-15
vim: Use `target` argument instead of `shorthand`

josephfrazier:
Commit 9750462 accidentally broke vimrc support, but this fixes it.

Test page: https://github.com/gmarik/vimfiles/blob/1f4f26d42f54443f1158e0009746a56b9a28b053/vimrc#L140-L144


stefanbuck:
:+1:

PR #349, opened 2017-04-17
Remove .hound.yml

josephfrazier:
See https://github.com/OctoLinker/browser-extension/pull/321#issuecomment-292153699


stefanbuck:
Thanks for taking care of it :+1:

Issue #351, opened 2017-04-24
Version 4.10.0 failed Mozilla review


josephfrazier:
From https://addons.mozilla.org/developers/addon/octolinker/versions:

> This version didn't pass review because of the following problems:
> 1) Please remove the `new Function` to eval code. You have it in the content script, which is not affected by any CSP.

I'm going to refer the reviewer to this issue, so we can track it and work on it publicly.

josephfrazier:
I did a `git bisect start master 4.8.0` and used the following commands to tell me whether a commit was "bad" or "good":

```bash
yarn
rm yarn.lock # if it isn't tracked in git at a given commit
yarn firefox-build
grep ' Function(' dist/app.js
```

I found that the number of instances of Function being used in this way has been increasing, but the first one appeared in commit e23ab38e07fb53128e3370a9d5c66734ffa5a4c6 (from [webpack/buildin/global.js](https://github.com/webpack/webpack/blob/34315db3ffe20b3d7e9a50885ea623fc90ac0861/buildin/global.js#L10)). The second one was added in 2545cae69e6166f72e345756c9105ca55ebc02b1 (from [core-js/modules/_global.js](https://github.com/zloirock/core-js/blob/f3ed6428fe0a182461acb844a5b2ab9bd8b414ac/modules/_global.js#L3)). The remaining instances were added in 642b4eb47bbcce59df2eb78f2415c539adb8c354 (from [core-js/modules/_global.js](https://github.com/zloirock/core-js/blob/f3ed6428fe0a182461acb844a5b2ab9bd8b414ac/modules/_global.js#L3) and [core-js/modules/_task.js](https://github.com/zloirock/core-js/blob/f3ed6428fe0a182461acb844a5b2ab9bd8b414ac/modules/_task.js#L32)).

josephfrazier:
I've merged #353, published v4.11.0, and manually submitted v4.11.0 to both [Mozilla](https://addons.mozilla.org/en-US/developers/addon/octolinker/versions) and [Opera](https://addons.opera.com/developer/package/226344/?tab=conversation). For what it's worth, v4.11.0 is currently at position 283 of 283 in the Mozilla review queue. I'll close this issue, and open a new one if v4.11.0 fails review.

PR #353, opened 2017-04-24
Inline babel helpers rather than requiring them

josephfrazier:
This addresses https://github.com/OctoLinker/browser-extension/issues/351
by removing a couple of `Function()` calls from `dist/app.js`. Here's
the diff in the output of `grep ' Function(' dist/app.js`:

```diff
diff --git a/before.txt b/after.txt
index cfc9de2..c5c2ee1 100644
--- a/before.txt
+++ b/after.txt
@@ -1,4 +1,2 @@
-  ? window : typeof self != 'undefined' && self.Math == Math ? self : Function('return this')();
 	g = g || Function("return this")() || (1,eval)("this");
-      invoke(typeof fn == 'function' ? fn : Function(fn), args);
   ? window : typeof self != 'undefined' && self.Math == Math ? self : Function('return this')();
```

Hopefully, the remaining two calls will be ok, since they simply get the
global object, rather than have the potential to execute arbitrary code.

EDIT: See here for more info about the `helpers` option: https://github.com/babel/babel/tree/412180e20304347f8c402f7e3c0613951ac5144c/packages/babel-plugin-transform-runtime#helpers

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
       [
         "transform-runtime",
         {
+          "helpers": false,
           "polyfill": false
         }
       ],


Noitidart:
Thanks for working with the review policies @josephfrazier !

josephfrazier:
Thank *you*, @Noitidart! For future reference, here is the Mozilla position on `eval`/`Function`, copied from our discussion on AMO:

> Yes, it is ok to `eval` or `new Function` with static strings. So `new Function('return this')` is ok but `let str = 'return this'; new Function(str)` is not.

PR #366, opened 2017-06-05
Fix .net core parsing for framework-specific dependencies

olsh:
The extension doesn't parse framework-specific dependencies in project.json file.
Here is an example of such file:
https://github.com/olsh/todoist-net/blob/master/src/Todoist.Net.Tests/project.json (lines 13 and 21)


josephfrazier:
Thanks @olsh!

PR #367, opened 2017-06-05
Add support for .NET Framework packages.config

olsh:
Here is an example of packages.config:
https://github.com/olsh/resharper-configuration-sense/blob/master/src/Resharper.ConfigurationSense/packages.config

### DIFF
--- a/lib/plugins/dot-net.js
+++ b/lib/plugins/dot-net.js
@@ -1,0 +1,37 @@
+import insertLink from '../insert-link';
+import nugetResolver from '../resolver/nuget';
+import { xmlRegExKeyValue } from './helper/regex-builder';
+
+function linkDependency(blob, key, value) {
+  const regex = xmlRegExKeyValue(key, value);
+
+  insertLink(blob.el, regex, {
+    pluginName: 'DotNet',
+    target: '$1',
+  });
+}
+
+export default {
+  name: 'DotNet',
+
+  resolve({ target }) {
+    return nugetResolver({ target });
+  },
+
+  getPattern() {
+    return {
+      pathPatterns: ['/packages.config$'],
+      githubClasses: [],
+    };
+  },
+
+  parseBlob(blob) {
+    const packageRegex = new RegExp('<\\s*package\\s.*id="(.*?)".*\\/>', 'g');
+    const blobString = blob.toString();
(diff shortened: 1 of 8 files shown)


lib/plugins/dot-net.js
```diff
@@ -0,0 +1,37 @@
...
+
+  getPattern() {
+    return {
+      pathPatterns: ['/packages.config$'],
+      githubClasses: [],
+    };
+  },
+
+  parseBlob(blob) {
+    const packageRegex = new RegExp('<\\s*package\\s.*id="(.*?)".*\\/>', 'g');
```
josephfrazier: Could you move this RegExp into [helper-grammar-regex-collection](https://github.com/OctoLinker/browser-extension/blob/02c0d0654a1436cdc90da33c92f82eb05463e8c2/packages/helper-grammar-regex-collection/index.js) and add a few [tests](https://github.com/OctoLinker/browser-extension/blob/02c0d0654a1436cdc90da33c92f82eb05463e8c2/packages/helper-grammar-regex-collection/test.js) of it? (commit a3ec0f9c048873509b6c60bd05e0019d2cf32657 has an example of adding/testing a new regex)

test/plugins/dot-net.test.js
```diff
@@ -0,0 +1,18 @@
...
+    );
+  });
+
+  it('resolves Stanford.NLP.CoreNLP to https://www.nuget.org/packages/Stanford.NLP.CoreNLP', () => {
+    assert.deepEqual(
+      dotNet.resolve({ target: 'Stanford.NLP.CoreNLP' }),
+      ['https://www.nuget.org/packages/Stanford.NLP.CoreNLP'],
+    );
+  });
+});
```
josephfrazier: Could you also add some fixtures here in order to make sure that the `NET_PACKAGE` regex matches the package name correctly? https://github.com/OctoLinker/browser-extension/blob/02c0d0654a1436cdc90da33c92f82eb05463e8c2/packages/helper-grammar-regex-collection/test.js#L6

josephfrazier:
Thanks, this work looks really good! I took the liberty of making one small tweak to the README before merging, rather than making you do it. Hope you don't mind :)

olsh:
@josephfrazier thanks! :)

PR #373, opened 2017-08-02
Change rust regex to be more exclusive

### LINKED ISSUE #371: Octolinker incorrectly identifies modules in Rust even when they are not specified
**Browser name:** 
Chrome

**Browser version:** 
Version 59.0.3071.115 (Official Build) (64-bit)

**OctoLinker version:** 
4.12.0

**URL and line number where issue occurs:** 
https://github.com/cynicaldevil/rust/commit/9713020dbe338e2a576b656d17f0415df26862a8#commitcomment-23137541

In this GH comment, Octolinker incorrectly assumes any word that comes after `use` as a module/package name.

cynicaldevil:
Fixes #371
cc @josephfrazier

### DIFF
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -77,7 +77,7 @@ export const VIM_PLUGIN = regex`
 `;
 
 export const RUST_CRATE = regex`
-  \b (extern \s crate | use)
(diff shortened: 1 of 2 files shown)


josephfrazier:
Thanks! I'm going to add a regression test to this later, then merge it in.

PR #378, opened 2017-10-04
Move to webpack css-loader

stefanbuck:
This PR will change the way how we load css. Before we make use of link tag to load css into the options page and a special manifest attribute to load css for the extension itself. To simplify this I added webpack`s css and style loader so we can treat it like any other imports.


stefanbuck:
Sure, it works 😉  Thanks for reviewing.

PR #380, opened 2017-10-05
Add Prettier

stefanbuck:
This pull request accomplishes 2 things

1. Adds Prettier and the standard configuration for it
2. Perform Prettier on all files

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,11 +1,15 @@
 module.exports = {
-  extends: ['airbnb-base', 'plugin:import/errors'],
-  plugins: ['mocha'],
+  extends: ['airbnb-base', 'plugin:import/errors', 'prettier'],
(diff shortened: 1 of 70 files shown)


josephfrazier:
This is great! I'm a big fan of Prettier, so it's nice to have it here :+1:

stefanbuck:
I wanted to have a final look before merging 😄 My fault, I should have used a label to mark this as not ready to merge. Anyway should be fine I guess

josephfrazier:
Oops, sorry about that! For what it's worth, I did try to verify that the source code AST didn't change by running `git diff stefanbuck/prettier^ stefanbuck/prettier --name-only | xargs prettier --debug-check && echo ok`. See here for more info on `prettier --debug-check`: https://github.com/prettier/prettier/tree/23f7e9216219da5b725524802522b70b5cef4f06#--debug-check

Issue #372, opened 2017-07-28
Missing links in golang.org/x/ imports


guneysus:
![image](https://user-images.githubusercontent.com/949232/28715976-6eef4246-73a3-11e7-81c9-218617ead61b.png)

Missing links in `golang.org/x` imports.

Example:

https://github.com/rs/xhandler#using-xmux

rgroothuijsen:
Still unsure about one thing. Would it be better to have the links pointing to the original documentation as they do now if you went to a golang.org package URL, or to have the links pointing to the Github mirrors where the actual code is located?

rgroothuijsen:
Small question: right now, the golang.org/x/ and golang.org/pkg/ URLs point to documentation pages for those packages. Would it be more helpful to have them redirect to the actual source code or keep the link to the documentation page?

stefanbuck:
That's just my personal view, but I think that pointing to the documentation is more useful. In most cases documentation pages are easier to access than raw source code and documentation is all what you need. We do the same for other languages like nodejs where we link core pages to the official documentation. I'm happy to discuss this further if you have strong feelings for your proposal.

rgroothuijsen:
Oh, not at all, I thought there was merit in both approaches so I was doubting which one to go for. If it's consistent with how other languages are handled, that's a plus. In that case, the original paths can be used without any special parsing and it's just a matter of adding golang.org to the list of allowed domains.

stefanbuck:
If you'd like to tackle this one, I would be happy to help with assistance (if needed). It will be fairly straightforward 😉

stefanbuck:
Fixed as part of #381 and will be shipped later this week

PR #381, opened 2017-10-07
Add golang.org to list of allowed domains for Go files

rgroothuijsen:
This PR addresses #372 by parsing golang.org imports. Since these import URLs already send the user to a documentation page when opened in a browser, no special parsing seems to be required.


stefanbuck:
Great 👍 Just one thing, please add a test [similar to this](https://github.com/octolinker/browser-extension/blob/5d0f537f942d15737d17c224accec96b23dd9ba5/packages/helper-grammar-regex-collection/test.js#L355). It's super easy 😉

rgroothuijsen:
Hmm, prettier complains about something, but the only difference I can see is some "⏎········" bits. Possible Linux vs. Windows CRLF issues?

stefanbuck:
Splitting into two lines sounds good to me. Thanks for addressing my comment.

stefanbuck:
I'm sorry, but there are still two linting issues left. You can easily fix them by running `npm run lint -- --fix`

rgroothuijsen:
Generally true, although since the branch was created before the introduction of prettier, I'll have to update it first.

stefanbuck:
Great, thanks

PR #383, opened 2017-10-10
Handle HTML script tags

### LINKED ISSUE #382: [Feature request] HTML script relative path linking
**Browser name:** Chrome

**Browser version:** 61.0.3163.100 (Official Build) (64-bit)

**OctoLinker version:**  4.12.0

**URL and line number where issue occurs:** N/A

**Expected behaviour:** OctoLinker does not or unable to link HTML script's **src** tag, like
`<script src="../components/temp/script.js"></script>`. You can see example [here](https://github.com/sumanmaity112/juice/blob/master/HTML/juicePiChart.html). Here you should be to click on `<script src="./js/juicePiChart.js"></script>` and see the file.

**Actual behaviour:** N/A

deecewan:
Closes #382 

Basic handling of `<script src="myfile.js"></script>`. See the new tests for a full breakdown of what I've handled.

While I was doing this, I noticed that absolute URL sources, such as on a CDN or similar, do not work with this. I can't see it being super useful, but what are your thoughts on putting something like:

```js
if (/^(https?:)?\/\//.test(target)) {
  return [target];
}
```

in the `resolve` function so that the absolute URLs open up where they target.

### DIFF
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -131,18 +131,12 @@ export const HTML_IMPORT = regex`
   href=${captureQuotedWord}>
 `;
 
-// handles the case the HTML script tag has a type with it
-const typeRegex = regex`
-  (\stype=['"][^'"\s]+['"])?
-`;
-
 export const HTML_SCRIPT_IMPORT = regex`
-  ^\s*
   <script
-  ${typeRegex}
+  [^>]"
   \s
   src=${captureQuotedWord}
-  ${typeRegex}
+  [^>]"
   \s*>
 `;
 
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -133,10 +133,10 @@ export const HTML_IMPORT = regex`
 
 export const HTML_SCRIPT_IMPORT = regex`
   <script
-  [^>]"
+  [^>]*
   \s
   src=${captureQuotedWord}
-  [^>]"
+  [^>]*
   \s*>
 `;
 
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -131,6 +131,21 @@ export const HTML_IMPORT = regex`
   href=${captureQuotedWord}>
 `;
 
+// handles the case the HTML script tag has a type with it
+const typeRegex = regex`
+  (\stype=['"][^'"\s]+['"])?
+`;
+
+export const HTML_SCRIPT_IMPORT = regex`
+  ^\s*
+  <script
+  ${typeRegex}
+  \s
+  src=${captureQuotedWord}
+  ${typeRegex}
+  \s*>
+`;
+
 export const NET_PACKAGE = regex`
   <\s*
   package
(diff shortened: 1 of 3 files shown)


packages/helper-grammar-regex-collection/index.js
```diff
@@ -131,6 +131,21 @@ export const HTML_IMPORT = regex`
...
+  (\stype=['"][^'"\s]+['"])?
+`;
+
+export const HTML_SCRIPT_IMPORT = regex`
+  ^\s*
+  <script
+  ${typeRegex}
+  \s
+  src=${captureQuotedWord}
+  ${typeRegex}
```
josephfrazier: Do we need `typeRegex` at all? I'm thinking we could simplify the code and support the presence of other attributes by replacing both instances of `${typeRegex}` with `[^>]*` so anything that's not the end of the tag is matched.

packages/helper-grammar-regex-collection/index.js
```diff
@@ -131,6 +131,21 @@ export const HTML_IMPORT = regex`
   href=${captureQuotedWord}>
 `;
 
+// handles the case the HTML script tag has a type with it
+const typeRegex = regex`
+  (\stype=['"][^'"\s]+['"])?
+`;
+
+export const HTML_SCRIPT_IMPORT = regex`
+  ^\s*
```
josephfrazier: I think we may want to delete this line, so that the regex matches `<script>` tags which aren't the first tag on their line.

packages/helper-grammar-regex-collection/index.js
deecewan: that's a really good point.

deecewan:
@josephfrazier thanks for the review. any thoughts on the absolute URL opening I mentioned in the opening?

stefanbuck:
First, thank you for the work 👍 Overall this looks good to me just few thing. You're assuming that a file which is referenced by a script tag is relative located to the HTML file. I don't have numbers, but I think in quite a lot of scenario this isn't true. For example if the HTML file is just a template file or a build step moves files around. I'm pretty sure this will come up sooner or later ... but let's keep it simple for now. 

Another thing which pops up in my head while writing those lines, why do we want do limit this the script tags. Wouldn't be that also useful for `<link>` tags? However, not important in scope of this PR 😉 

Regarding the issue you mentioned about absolute urls, this is something that I wanted to implement long time ago. However since the [refine github](https://github.com/sindresorhus/refined-github) extension does it, I wasn't in the need of adding this functionally. Maybe we should add it, because not everyone has refined-github installed.

deecewan:
@stefanbuck thanks for the feedback.

I agree that a lot of the scripts here will be './dist/index.js', where './dist' doesn't actually exist before the build step. Unfortunately I'm not sure that there is too much that can be done about that 😢 .

As for links, the regex right about the regex I added already does this 😄 .

Yup. I use refined-github, so I didn't notice it until I was testing out my changes last night at the url linked in #382, which contains a absolute URL. I think that it's also probably outside the scope of this PR, but is definitely food for thought.

stefanbuck:
There's also a resolver which wraps the github search API which could be used to link beyond relative files https://github.com/OctoLinker/browser-extension/blob/master/lib/resolver/github-search.js I forgot to mention that. We do this for ruby files where it's hard to predict the location of a file.

stefanbuck:
I'm gonna to merge this now and we can do further improvements separately. A new version of OctoLinker will be available probably next week

PR #386, opened 2017-10-16
Refactor options page 

stefanbuck:
The current architecture of the options page isn't that great. I refactored it to Preact, which feels much cleaner and modern now. 

@josephfrazier I already did a quick QA check on this and will do another one tomorrow evening. Would be great if you could also verify that nothing breaks.


stefanbuck:
I was referencing the wrong field for the "New Tab" options. It's fixed now and I did another QA round and I didn't noticed any issues.

.eslintrc.js
```diff
@@ -1,15 +1,29 @@
 module.exports = {
-  extends: ['airbnb-base', 'plugin:import/errors', 'prettier'],
-  plugins: ['prettier', 'mocha'],
+  parser: 'babel-eslint',
+  extends: ['airbnb', 'prettier'],
```
josephfrazier: Do we need to remove `plugin:import/errors` here?

stefanbuck: I was comparing the eslint config with another project and there I didn't use the `eslint-plugin-import`, so I thought let's remove it. Fun fact: I forgot to remove the dependency which I did just right now. However now I get this error `couldn't find the plugin "eslint-plugin-import".` and I don't know where this error comes from. 

For now I think it's okay to keep it like it is and I'll revisit this some day. It shouldn't harm even the plugin is magically in use.

stefanbuck:
Appveyor's branch status is pending since yesterday. A restart didn't fixed it. I'll merge it and let's hope the best

.eslintrc.js
josephfrazier: It looks like `eslint-config-airbnb-base` includes `eslint-plugin-import`, so we're still implicitly using it: https://github.com/airbnb/javascript/blob/ad3746672e5bd79f77159c4e2612077651e9af17/packages/eslint-config-airbnb-base/package.json#L56

stefanbuck: Hold on, I removed `eslint-config-airbnb-base` https://github.com/OctoLinker/browser-extension/pull/386/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2, but I fear I didn't used `yarn remove *`. Will fix this in another PR

stefanbuck: Looks like everything is correct. `eslint-config-airbnb` includes `eslint-config-airbnb-base` which is the reason why it's still listed in `yarn.lock`. Anyway, let's ignore this for now 😉

josephfrazier: Yes, I think everything is in order, and there's no issue remaining to ignore :)

lib/options/page.js
```diff
@@ -1,28 +1,80 @@
-import 'primer-core/build/build.css';
-import 'primer-forms/build/build.css';
-import './options.css';
+import { h, render, Component } from 'preact';
+import linkState from 'linkstate';
 
-import components from './components';
+import './options.css';
+import { Input, Checkbox } from './components';
```
stefanbuck: Just a test

PR #393, opened 2017-11-08
Drop core-js

bfred-it:
It's been supported for a while now, I'd argue that most babel transforms you have there are unnecessary. I think you probably exclusively need the JSX transform.

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -8,9 +8,11 @@
   "icons": {
     "128": "icon.png"
   },
+  "minimum_chrome_version": "54",
   "applications": {
     "gecko": {
-      "id": "octolinker@stefanbuck.com"
+      "id": "octolinker@stefanbuck.com",
+      "strict_min_version": "47.0"
     }
   },
   "background": {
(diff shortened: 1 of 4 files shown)


assets/manifest.json
```diff
@@ -8,9 +8,11 @@
   "icons": {
     "128": "icon.png"
   },
+  "minimum_chrome_version": "54",
```
stefanbuck: I think this will make the manifest incompatible with Firefox. Would be great if you could quickly verify this by running `yarn firefox-build && yarn firefox-launch`

stefanbuck: If yes, you can remove easily remove the field during the build like we do for `applications` here https://github.com/octolinker/browser-extension/blob/6ca6f7604de0847021e8caf91a66f170284c549e/package.json#L16

bfred-it: No, Firefox doesn't complain about that key. I mean even Chrome doesn't really care about it, it only shows the warning in the development version, but it still works fine. In my extensions I just load the same `manifest.json` in both.

stefanbuck: Ahh cool, I didn't know that. Will update the build step one day

PR #394, opened 2017-11-08
Drop most babel transforms

bfred-it:
As discussed in #393


bfred-it:
Why are the tests failing? It doesn't show any error

josephfrazier:
It looks like both [Chrome](https://travis-ci.org/OctoLinker/browser-extension/builds/298933583#L731-L733) and [Firefox](https://travis-ci.org/OctoLinker/browser-extension/builds/298933583#L735-L737) have some issue with assigning to `module.exports`.

bfred-it:
What’s that line doing anyway? It looks like it runs a loop on itself

stefanbuck:
That's a [dynamic require](https://webpack.js.org/guides/dependency-management/#require-context) which is needed to load the tests files. Not sure why your change is conflicting with reassigning of  `module.exports`, but I noticed that we can remove this line anyway. Also in the `main-packages.js`which loads the test from `/packages` doesn't need this export assignment.

bfred-it:
Can you apply that change to this PR?

stefanbuck:
It seems that the rebase has some effect on the build 🤷

josephfrazier:
Huh, that's interesting. I guess it had something to do with `core-js`: https://github.com/OctoLinker/browser-extension/commit/e56f75d5b5f22bb5d27ed46775d0d056d19600b2#diff-e00d3d45185f722f3df476b178fdd069

PR #395, opened 2017-11-08
Drop browser-specific packages

bfred-it:
As discussed in #393


stefanbuck:
Well done. Many thanks

PR #398, opened 2017-11-10
Add support for .NET project files

xt0rted:
This adds support for NuGet references in csproj & vbproj files. This is the newer replacement for the .net core project.json setup (which there's existing support for), and is also usable in older projects in VS 2017.

`PackageReference` docs are located at https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files

I based this on the work done in #367 but if there's anything that needs changing just say the word.

When I ran `yarn lint` locally I got a ton of line ending errors (I'm on Windows and git's setup to checkout with CrLf and commit with Lf). Hopefully there's no issues and CI passes 🤞

Some examples of where this *should* work are

**csproj**
https://github.com/aspnet/Mvc/blob/a01e23250f98fbce016cd2b3bdf69e43fc4505c8/src/Microsoft.AspNetCore.Mvc/Microsoft.AspNetCore.Mvc.csproj#L21-L22

https://github.com/brandonseydel/MailChimp.Net/blob/00cae16a69d5a8779093b37a845aa686b4def66a/MailChimp.Net/MailChimp.Net.csproj#L52-L58

**vbproj**
https://github.com/dotnet/roslyn/blob/a604cefecd2cba10803a872c580cfb7a9e0883ad/src/Compilers/VisualBasic/Test/Syntax/BasicCompilerSyntaxTest.vbproj#L34-L36

There is a minor bug in this, but I don't think it's due to the changes I made. If you run this version and look at the vbproj file above the first reference results in 2 links being inserted. I'm guessing this is because there's two instances of `xunit` on that line. I'm not familiar with the inner workings of this plugin but if someone could point me to where that might be coming from I can take a look at it.

The resulting html is
```html
<td id="LC34" class="blob-code blob-code-inner js-file-line highlighted"><span class="glebm-ws glebm-ws-s-4">    </span>&lt;<span class="pl-ent">PackageReference</span><span class="glebm-ws glebm-ws-s-1"> </span><span class="pl-e">Include</span>=<span class="pl-s"><span class="pl-pds">"</span><span><a class="octolinker-link octolinker-line-indicator" data-plugin-name="DotNetProj" data-target="xunit" data-path="/dotnet/roslyn/blob/a604cefecd2cba10803a872c580cfb7a9e0883ad/src/Compilers/VisualBasic/Test/Syntax/BasicCompilerSyntaxTest.vbproj"><span>xunit</span></a></span><span class="pl-pds">"</span></span><span class="glebm-ws glebm-ws-s-1"> </span><span class="pl-e">Version</span>=<span class="pl-s"><span class="pl-pds">"</span><span>$(<a class="octolinker-link octolinker-line-indicator" data-plugin-name="DotNetProj" data-target="xunit" data-path="/dotnet/roslyn/blob/a604cefecd2cba10803a872c580cfb7a9e0883ad/src/Compilers/VisualBasic/Test/Syntax/BasicCompilerSyntaxTest.vbproj"><span>xunit</span></a>Version)</span><span class="pl-pds">"</span></span><span class="glebm-ws glebm-ws-s-1"> </span>/&gt;</td>
```


packages/helper-grammar-regex-collection/test.js
```diff
@@ -400,6 +400,18 @@ const fixtures = {
       'import com.company..myapp',
     ],
   },
+  NET_PROJ_PACKAGE: {
+    valid: [
+      '<PackageReference Include="EntityFramework" Version="6.2.0" />',
+      '< PackageReference Include="EntityFramework" Version="6.2.0" />',
```
stefanbuck: Are you sure that a space in front of `PackageReference ` is a valid format?

packages/helper-grammar-regex-collection/test.js
```diff
@@ -400,6 +400,18 @@ const fixtures = {
...
+      '<PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="$(MicrosoftExtensionsCachingMemoryPackageVersion)" />',
+      '<PackageReference Include="Microsoft.AspNetCore.Cors">\n<Version>2.0.0</Version>\n</PackageReference>',
```
stefanbuck: Is the order of the attributes always the same or should be the Regex also allow  `<PackageReference Version="6.2.0" Include="EntityFramework" />`

packages/helper-grammar-regex-collection/test.js
stefanbuck: Did you run `yarn test` locally? I think your tests will fail. By default the test suite which is made out of this config expects "foo" as target.

You could either change the lines to

```js
<PackageReference Include="foo" Version="6.2.0" />
```
or use an array where the second item descibes the expected result.

```js
['<PackageReference Include="EntityFramework" Version="6.2.0" />','EntityFramework']
```

xt0rted: As far as I know a space beween `<` and the tag name isn't valid, but I wasn't sure if github may add space when rendering the html (the `NET_PACKAGE` regex had it, so I followed how that was setup). I've removed it though since it doesn't seem right and we can always adjust it later if there's situations where it may be needed.

packages/helper-grammar-regex-collection/test.js
xt0rted: Good point, I've adjusted the regex so order doesn't matter and added a test for it

xt0rted:
Running `yarn lint --fix` did fix the errors, but now I have 92 changed files that want to be committed and all that's changed in them is their line endings.

lib/plugins/dot-net.js
```diff
@@ -1,5 +1,8 @@
 import nugetResolver from '../resolver/nuget';
-import { NET_PACKAGE } from '../../packages/helper-grammar-regex-collection/index.js';
+import {
+  NET_PACKAGE,
+  NET_PROJ_PACKAGE,
```
xt0rted: This is the format `yarn lint` wanted, even though in other files they're on one line, such as in https://github.com/OctoLinker/browser-extension/blob/d781585917cbca8c86e43b73222ccc6e29114375/lib/plugins/bower-manifest.js#L4

xt0rted:
It might be helpful to add a comment about line endings on windows & using `yarn lint --fix` or possibly forcing the repo to checkout with Lf instead of CrLf. It's never fun seeing a bunch of errors upon checkout.

It could also be useful to put a note in [helper-grammar-regex-collection/README.md](https://github.com/OctoLinker/browser-extension/blob/master/packages/helper-grammar-regex-collection/README.md) that the tests expect `foo` in the valid results. I didn't realize that was the case just by looking at the file. Once you mentioned it it made sense, but before that it just looked like basic test data.

Aside from that it was pretty easy to find what I needed for the initial changes. In fact it was so easy I just made the PR instead of filing a bug hoping someone else would do it. 

This is the minor bug I  mentioned. I've looked through the code more but I'm not sure where this is coming from or if it's even an easy fix. The second `xunit` link shouldn't be a link but it is.

image

stefanbuck:
Thanks for your feedback. @josephfrazier raised a PR to enforce to correct line endings #399 so this shouldn't be an issue anymore. 

Regarding the `helper-grammar-regex-collection` tests. I think we should get rid of the `foo` magic and use the array pattern `[input, expect]` for all tests. 

Later this week I'll try to find some time to have a closer look at bug that you have found. Most likely the bug sit in [insert-link.js](https://github.com/OctoLinker/browser-extension/blob/master/lib/insert-link.js) which is the most tricky part of this repo 😅

josephfrazier:
> Regarding the `helper-grammar-regex-collection` tests. I think we should get rid of the `foo` magic and use the array pattern `[input, expect]` for all tests. 

I'll try to take a shot at this. Hoping it'll be relatively straightforward with a little editor scripting.

PR #400, opened 2017-11-15
helper-grammar-regex-collection: Explicitly specify 'foo' for valid matches

josephfrazier:
Addresses https://github.com/OctoLinker/browser-extension/pull/398#issuecomment-344412479

It's easier to see the changes in this commit with:

    git show --color-words=.


packages/helper-grammar-regex-collection/test.js
```diff
@@ -418,7 +436,7 @@ const fixtures = {
 function fixturesIterator(fixturesList, next) {
   fixturesList.forEach(statement => {
     const text = Array.isArray(statement) ? statement[0] : statement;
-    const expected = Array.isArray(statement) ? statement[1] : ['foo'];
+    const expected = Array.isArray(statement) ? statement[1] : null;
```
stefanbuck: Maybe I'm missing something here, but do we still need the ternary operator here? It should be always an array, right? 

```js
const [text, expected] = statement;
```

stefanbuck: Maybe I'm missing something here, but do we still need the ternary operator here? It should be always an array, right? 

```js
const [text, expected] = statement;
```

josephfrazier: The `invalid` test cases are still just a string, because I thought it seemed unnecessarily verbose to have to specify `null` in every `invalid` case. If you think it would be better to do that, however, I'd be happy to.

stefanbuck: I totaly forgot about the `invalid` test case 😉 Thanks for clarifying

PR #401, opened 2017-11-15
Update yarn.lock and have Travis enforce it

josephfrazier:
The following commits removed dependencies from package.json, but didn't
update yarn.lock:

* 5c8c1b19659c65f63782644efb95e6ea005a7aa3: Drop core-js (#393)
* d9038106635ee28ad2cef106387c7515646be532: Drop most babel transforms (#394)

This change updates yarn.lock, and also adds a bit to .travis.yml that
fails the build if yarn.lock changed after running `yarn`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.1.0
   - export PATH=$HOME/.yarn/bin:$PATH
 before_script:
+  - git diff --exit-code # make sure that yarn.lock didn't change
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -18,6 +18,7 @@ before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.1.0
   - export PATH=$HOME/.yarn/bin:$PATH
 before_script:
+  - git diff --exit-code # make sure that yarn.lock didn't change
```
stefanbuck: Wow ... that's super helpful 👍

josephfrazier: Thanks!

PR #403, opened 2017-11-27
Support require calls using backtick-quoted strings in JS

TehShrike:
Anywhere that a string literal can be used to import files in JS, backticks are valid:

```js
const someThing = require(`aw-yeah.js`)
```

This does not include ES Module imports/exports - those are restricted to single/double quotes.

I looked around for some tests to modify so I could add extra test cases, but I didn't see any tests that covered this - the tests in `test/grammar/` aren't being executed, and it looks like they were written for an older version of the library anyway.

### DIFF
--- a/packages/helper-grammar-regex-collection/test.js
+++ b/packages/helper-grammar-regex-collection/test.js
@@ -43,6 +43,7 @@ const fixtures = {
     ],
     invalid: [
       'import "fo o"',
+      'import foo from `foo`',
       // TODO tweak IMPORT regexp so that invalid statements are not matched
       // 'import foo "foo"',
       // 'import from "foo"',
@@ -58,7 +59,10 @@ const fixtures = {
       ['export { bar\n } from "foo"', ['foo']],
       ['export { \nbar\n } from "foo"', ['foo']],
     ],
-    invalid: ['export * from "fo o"'],
+    invalid: [
+      'export * from "fo o"',
+      'export * from `foo`',
+    ],
   },
   REQUIRE: {
     valid: [
@@ -79,6 +83,7 @@ const fixtures = {
       ['foo = require("foo")bar = require("bar")', ['foo', 'bar']],
       ['foo = require("a-b")bar = require("c-d-e")', ['a-b', 'c-d-e']],
       ['foo = require("./foo")bar = require("./bar")', ['./foo', './bar']],
+      ['foo = require(`./foo`)bar = require(`./bar`)', ['./foo', './bar']],
       [
         'const foo = require("./foo")bar = require("./bar")',
         ['./foo', './bar'],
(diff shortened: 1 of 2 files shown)


TehShrike:
The tests passed locally... I'm not sure what's up with those failures? :-x

TehShrike:
oh derp, I see.  The tests aren't in the `test` folder, they're in `packages/helper-grammar-regex-collection/test.js`.  I'll add some assertions.

packages/helper-grammar-regex-collection/test.js
```diff
@@ -120,7 +123,8 @@ const fixtures = {
       ['var _ = import("foo")', ['foo']],
       ['var foo = import("foo")var bar = import("bar")', ['foo', 'bar']],
       ['import("foo")import("bar");', ['foo', 'bar']],
-      ['import("foo")import("bar");', ['foo', 'bar']],
+      ['import("foo")import("bar")', ['foo', 'bar']],
```
josephfrazier: Can you change this line back to how it was before? I'm assuming it was an accidental change.

TehShrike: Sorry, I should have made a comment on this line.

I saw that this line was identical to the line above it, and guessed that it was meant to be like lines 128 and 129, where the first line had a semicolon and the second did not.

My bad for just changing it in this pull request.  Would you like the change rolled back, line 126 removed entirely, or the change left to avoid having two identical assertions?

josephfrazier: Oh, I hadn't even noticed that, thanks for explaining! I think it's fine to leave it as it currently.

TehShrike:
👍 

What's the publish cycle like - any guess at when a release could happen?

josephfrazier:
There's not really a well-defined schedule, it's more when @stefanbuck or I have the time to work through the [Release Checklist](https://github.com/OctoLinker/browser-extension/blob/67ea1825a87743e79a77a6da32cedacddf2b2fb4/CONTRIBUTING.md#release-checklist).

If you're using Chrome, you can install a locally built version as described in the [Chrome docs](https://developer.chrome.com/extensions/getstarted#unpacked). Opera may have a similar process, but I think Firefox does not, unless you're using a nightly/developer version and are willing to jump through some extra hoops (see the [Firefox docs](https://wiki.mozilla.org/Add-ons/Extension_Signing) for details on that).

TehShrike:
I recently changed the [`quotes`](https://eslint.org/docs/rules/quotes) setting in my eslint autoformat script to `backtick`, inspired in part by [this ponyfoo post](https://ponyfoo.com/articles/template-literals-strictly-better-strings), and in other part by the fact that it eliminates having to escape quotes in strings virtually all of the time. 😅

So it'll be spreading through [my projects](https://www.npmjs.com/~tehshrike) and apps at least, but as you can see from that year-old ponyfoo post, I'm not the first.

josephfrazier:
Thanks for the context, that makes a lot of sense!

stefanbuck:
@TehShrike I was off sick this week and therefore I didn't had time to finish the stuff that I would like to have in the upcoming release as well. I'm aiming to finish them over the weekend and release a new version beginning of next week.

PR #409, opened 2017-12-04
Take addition and deletion sign into account when building regex

stefanbuck:
Today I noticed that a few plugins doesn't work for deleted and added lines in the PR diff view. A deletion starts with a minus sign and an addition with a plus sign. Some RegExp exclude those lines by using the `^` followed by the keyword expression. This PR tweak them to take the diff signs into account. 

@josephfrazier I only updated the RegExp which already started with the `^` anchor. Maybe we should update the other as well. What do you think?

### DIFF
--- a/packages/helper-grammar-regex-collection/test.js
+++ b/packages/helper-grammar-regex-collection/test.js
@@ -459,9 +459,20 @@ function fixturesIterator(fixturesList, next) {
   });
 }
 
+function addModifiedLines(valid) {
+  const [text, expected] = valid[0];
+  const diffLines = [[`-${text}`, expected], [`+${text}`, expected]];
+
+  return [].concat([], diffLines, valid);
+}
+
 describe('helper-grammar-regex-collection', () => {
   Object.keys(fixtures).forEach(grammar => {
-    const { valid, invalid } = fixtures[grammar];
+    const spec = fixtures[grammar];
+
+    const { invalid } = spec;
+    const valid = addModifiedLines(spec.valid);
+
     let regexes = REGEX[grammar];
 
     // Help ensure that `regexes` is a function that returns an array of RegExp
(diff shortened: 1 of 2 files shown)


packages/helper-grammar-regex-collection/test.js
```diff
@@ -459,9 +459,20 @@ function fixturesIterator(fixturesList, next) {
   });
 }
 
+function addModifiedLines(valid) {
+  const [text, expected] = valid[0];
+  const diffLines = [[`-${text}`, expected], [`+${text}`, expected]];
+
+  return [].concat([], diffLines, valid);
+}
```
stefanbuck: This automatically adds test for deletion and additions. I hope no one will be confused by this, but it feels worng to me to add those to every single test manually.

josephfrazier:
Oops, completely forgot about this one... It looks good to me!

stefanbuck:
No worries, what do you think about the other tests? Should we prefix all of them with this RegExp pattern?

josephfrazier:
Oh, because [addModifiedLines](https://github.com/OctoLinker/browser-extension/blob/07dfc9f55814ad664974a8a4ddb0858d56c75d99/packages/helper-grammar-regex-collection/test.js#L462) only does it for the first test? Yeah, it seems reasonable to do it for every test.

stefanbuck:
It was already too late when I was writing the comment and a wrong word sneaked in. I meant the other regexp not tests. Sorry, miscommunication from my side. 

Basically I was wondering if we have to apply this change that I made to all existing RegExp. 

For example the IMPORT regex is very generic

https://github.com/OctoLinker/browser-extension/blob/07dfc9f55814ad664974a8a4ddb0858d56c75d99/packages/helper-grammar-regex-collection/index.js#L44-L48

Therefore it will also match

```
some-other-string-or-character import "./file.js"
```

which is good for increasing the coverage rate, but not very precise – I'm undecided.

PR #410, opened 2017-12-06
Handle leading whitespace in diff line

stefanbuck:
#409 introduced a bug that stopped OctoLinker from inserting links. On GitHub a diff line starts with a whitespace character if the line is not modified which the RegeEx doesn't take into account. 

### Before
screen shot 2017-12-06 at 23 44 30

### After
screen shot 2017-12-06 at 23 43 48

### DIFF
--- a/packages/helper-grammar-regex-collection/test.js
+++ b/packages/helper-grammar-regex-collection/test.js
@@ -461,7 +461,7 @@ function fixturesIterator(fixturesList, next) {
 
 function addModifiedLines(valid) {
   const [text, expected] = valid[0];
-  const diffLines = [[`-${text}`, expected], [`+${text}`, expected]];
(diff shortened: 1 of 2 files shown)


packages/helper-grammar-regex-collection/test.js
```diff
@@ -461,7 +461,7 @@ function fixturesIterator(fixturesList, next) {
 
 function addModifiedLines(valid) {
   const [text, expected] = valid[0];
-  const diffLines = [[`-${text}`, expected], [`+${text}`, expected]];
+  const diffLines = [[` -${text}`, expected], [` +${text}`, expected]];
```
stefanbuck: @josephfrazier this small change is testing the new behaviour

Issue #418, opened 2017-12-27
Unable to build extension


xt0rted:
I'm working on a new PR and since my last one something's changed and I'm no longer able to build. Before I started on this change I deleted my local repo and recloned from GitHub so the recent line ending change worked.

I can run `yarn install`, `yarn lint` and `yarn test` just fine, but `yarn build`, `yarn watch`, and `yarn chrome-open` all return errors.

I'm on Win10 and tried with both the regular console and powershell but both gave the same results.

I think https://github.com/OctoLinker/browser-extension/commit/e50f33972621e47af6e4f56afea6ac22128af3ea#diff-63ac5d48d83927921b7bcdb2b5a8d124 might be what broke this.

```
C:\dev\GitHub\browser-extension [master ≡ +0 ~2 -0 !]> yarn chrome-open
yarn run v1.3.2
$ npm run build && npm run chrome-launch --

> OctoLinker@4.15.1 build C:\dev\GitHub\browser-extension
> webpack

Hash: 7ce29e98274dd109223f
Version: webpack 3.6.0
Time: 2162ms
            Asset       Size  Chunks                    Chunk Names
           app.js     599 kB       0  [emitted]  [big]  app
       options.js     183 kB       1  [emitted]         options
    background.js    25.8 kB       2  [emitted]         background
       app.js.map     759 kB       0  [emitted]         app
   options.js.map     231 kB       1  [emitted]         options
background.js.map    32.6 kB       2  [emitted]         background
    manifest.json  935 bytes          [emitted]
     options.html  134 bytes          [emitted]
         icon.png    13.9 kB          [emitted]
   [1] ./lib/options/storage.js 1.31 kB {0} {1} {2} [built]
  [20] ./package.json 2.56 kB {0} {2} [built]
  [34] ./lib/app.js 204 bytes {0} [built]
  [35] ./lib/octo-linker.js 1.42 kB {0} [built]
  [38] ./packages/blob-reader/index.js 307 bytes {0} [built]
  [40] ./lib/click-handler.js 3.08 kB {0} [built]
  [42] ./lib/plugin-manager.js 2.08 kB {0} [built]
  [46] ./lib/load-plugins.js 1.4 kB {0} [built]
  [89] ./lib/app.css 1 kB {0} [built]
  [91] ./lib/background/index.js 235 bytes {2} [built]
  [92] ./lib/background/newTab.js 305 bytes {2} [built]
[...]

PR #419, opened 2017-12-27
Add cli tool support to .net proj files

xt0rted:
This expands the existing .net proj file linking to also support [cli references](https://docs.microsoft.com/en-us/dotnet/core/tools/extensibility) via `DotNetCliToolReference`.

The tests are passing locally, but as per #418 I'm unable to build & test it myself 😞 

An example of where this would work is https://github.com/aspnet/DotNetTools/blob/3c20e7c45beb5d45547b549964a3da83b2e8303e/samples/dotnet-watch/LaunchAnyCommand/LaunchAnyCommand.csproj#L6-L10

### DIFF
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -194,7 +194,7 @@ export const JAVA_IMPORT = regex`
 `;
 
 export const NET_PROJ_PACKAGE = regex`
-  <PackageReference
+  <(PackageReference|DotNetCliToolReference)
   \s+
   .*
   Include=${captureQuotedWord}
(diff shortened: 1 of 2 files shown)


xt0rted:
It looks like the same error/warning is happening on Travis & AppVeyor, but those builds continue on instead of erroring out like my local build.

xt0rted:
`yarn build` works now and the error/warning is gone. `yarn chrome-open` is still broke but that's a separate issue.

josephfrazier:
This looks great! Thanks for including a link to the relevant documentation, as well as an example URL. Those made it really easy to understand/test.

PR #421, opened 2017-12-28
Fix load plugins error

xt0rted:
This was introduced in #415 and is the root cause of the build error in #418.


josephfrazier:
Thanks, I'll merge as soon as CI passes!

Issue #432, opened 2018-01-12
Complete mono repo work


stefanbuck:
Long time ago I added the `packages` directory inspired by babels mono repo. Back then, I had some issues with getting [lerna](https://github.com/lerna/lerna) working. I would like to give it another try. 

The `lib` folder needs to go away in favor `packages/octolinker-core`. This comes with another 
advantage for the repo structure. All app specific dependencies needs to be defined within `packages/..`. This means the `package.json` file in the root will contain tooling related dependencies only. Over time we can refactor more and more stuff out of the `octolinker-core` into sensible small units.

Along those change, I also would like to rename the repo from `browser-extension` to `OctoLinker`.

What do you think @josephfrazier?

josephfrazier:
That sounds ok to me. Would you be interested in bringing `live-resolver` into this repo too? I'd love to be able to have a single `OctoLinker` repository that encompasses both client and server side code. For reference, we had briefly discussed this starting in https://github.com/OctoLinker/browser-extension/pull/87#issuecomment-222865992. I think this is the most relevant part of that thread:

> When I suggested combining the live-resolver into this repo, I didn't mean that all of the resolution for external packages should be done client-side (though that is an interesting idea). Instead, what I should have explained is that since the code is closely coupled across repos, it might make sense to combine them into one repo. We could still deploy the extension and server separately, but if there were to be any shared logic, it would be easier to reuse code for it than if the extension code and the server code were in separate repos.

stefanbuck:
Yeah let's do this. I opened a PR to bring lerna into the code base. Would you be up for adding the live-resolver whenever you have time? That would be helpful so I can focus on the real link work.

josephfrazier:
Yeah, once we have lerna working with the existing repo (I'm about to start reviewing #433), I'd be happy to look at adding the live-resolver. I imagine the trickiest part might be getting the heroku deploy working inside the same travis job as the client-side tests/deployment, but hopefully it won't be too bad.

stefanbuck:
Today, I was thinking about that as well, but yeah let's assume it's somehow possible.

stefanbuck:
I think we should keep the live-resolver in a separate repository so we don't have to deal with unnecessary issues.

stefanbuck:
@josephfrazier if you have another view on this, feel free to re-open this issue

PR #433, opened 2018-01-13
Add Lerna

stefanbuck:
As described in #432 this PR adds [lerna](https://github.com/lerna/lerna). A tool to manage multiple packages in the same repo. I moved the `lib` folder into a new package called `core` along with all dependencies. Also I added a few more `package.json` files and cleaned up the dependencies in the root.


stefanbuck:
Thanks for reviewing. I've created the commits with the review in mind. Sure I'll squash them as always ... now.

PR #435, opened 2018-01-15
Move insert link to 

stefanbuck:
The primary intent for this PR is moving `lib/insert-link.js` into a separate package. Except one change everything will remain the same. I removed the `link indicator` options mainly to avoid more refactoring 😜 I think there is no need for such an option.


packages/helper-insert-link/style.css
```diff
@@ -1,21 +1,13 @@
...
 }
 
 .octolinker-link:hover {
   text-decoration: underline !important;
 }
 
-body.octolinker-debug .octolinker-link {
-  background-color: rgba(255, 0, 255, 0.2);
-  border: 1px solid rgb(255, 0, 255);
-}
```
stefanbuck: Unused css

josephfrazier:
This seems mostly fine to me! I have a couple questions, though:

* Should we just call the new package `@octolinker/insert-link`, or did you make it `@octolinker/helper-insert-link` in order to be consistent with `@octolinker/helper-grammar-regex-collection`? Now that we have the `@octolinker` namespace, I think `helper` is a bit redundant/lengthy.
* Does removing the `link indicator` option actually save a lot of effort? It's technically a breaking change, so I'd slightly prefer to leave it in if we can.

stefanbuck:
Thanks for the review, really appreciated!  My thinking about the naming was, that everything which is used by the `core` package and can't be reused outside of this project goes into the `helper-*` namespace. On the other hand packages like`blob-reader` can be reused for other extensions as well so no `helper-` prefix for those. No rule without exception, plugins I would like to name just`plugins-*`. Does this make sense for you?

Regarding the option, you're right it's a breaking change, even I doubt that no one has ever disabled it. Maybe this is something for the next major release. Talking about that, I pushed a v5 branch, which I will start soon feeding with PRs around the `real-link` topic. However, I'll have a look how much effort is needed to make the option working again.

stefanbuck:
@josephfrazier I reverted some of the changes and added a new package `helper-settings`.

stefanbuck:
@josephfrazier are you happy with me merging this? I addressed your comment regarding the `link indicator` option and explained the naming convention. Let me know if there is anything else left.

josephfrazier:
Thanks for the reminder! Sorry, this and https://github.com/OctoLinker/OctoLinker/pull/436 had fallen off my radar again. The naming convention makes sense to me, and I'd definitely be open to removing the `link indicator` option later/separately.

stefanbuck:
No worries, I thought you're busy with other stuff or just off for a while.

PR #438, opened 2018-01-23
Support tsx

bildja:


josephfrazier:
Hi @bildja, thanks for the PR! I think tsx is already supported though, by [this line](https://github.com/OctoLinker/OctoLinker/blob/d81b97c4d108d10260ec084112e914430c261dce/packages/core/plugins/javascript.js#L23). Here's a page you can test it on: https://github.com/Microsoft/TypeScript-React-Starter/blob/68f60e1a4b947df47418e1d420acc59dafdfef12/src/index.tsx

Hmm... after looking a little more closely, only the typescript plugin matches the `TYPESCRIPT_REFERENCE` pattern, so this PR would make that pattern become a link for .tsx files, whereas it wouldn't before, so I'll go ahead and merge. However, maybe we should consider merging the typescript plugin into the javascript plugin, so that there's no overlap.

PR #441, opened 2018-02-01
Continue mono repo work for plugins and resolvers

stefanbuck:
This PR moves all plugins and resolvers from the core package into own packages. I wrote a [little script](https://gist.github.com/stefanbuck/9ead57ec64d1c00f91759183deed48ae) for moving the files around and based on there imports it generated a package.json. It's super hackish but worked quite well


josephfrazier:
:+1: It's awesome that you were able to do this programmatically!

Issue #443, opened 2018-02-07
Composer support broken


jdreesen:
**Browser name:** Firefox

**Browser version:** 58.0.1 (64-Bit)

**OctoLinker version:** 4.16.0

**URL and line number where issue occurs:** 
right now: https://github.com/matthiasnoback/behat-local-code-coverage-extension/blob/master/composer.json#L6
but it occurs on every `composer.json` file since a while

**Expected behavior:** A new tab for the package opens

**Actual behavior:** Nothing happens

![grafik](https://user-images.githubusercontent.com/424602/35940828-b5870dec-0c50-11e8-8156-86f776342d99.png)

stefanbuck:
Looks like my giant refactoring #436 cause this issue, sorry for that. Will fix it asap.

jdreesen:
Thx for the quick fix! I absolutely love this addon :)

stefanbuck:
New version `v4.16.1` is out.

jdreesen:
I just updated and it works again, thx ❤️

PR #444, opened 2018-02-07
Handle undefined results in the match array

stefanbuck:
Fix for #443. This was caused by combination, where the [composer plugin](https://github.com/OctoLinker/OctoLinker/blob/b175d59f1bd6f630ffb655dac6449a674852aa7f/packages/plugin-composer-manifest/index.js#L8) returns `undefined` if the json key is `php` and a not solid `forEach` loop in the octolinker core.

screen shot 2018-02-07 at 22 54 31

### DIFF
--- a/packages/core/octo-linker.js
+++ b/packages/core/octo-linker.js
@@ -18,7 +18,12 @@ function initialize(self) {
 }
 
 function insertDataAttr(matches) {
-  matches.forEach(({ data, link }) => {
+  matches.forEach(item => {
+    if (!item) {
+      return;
+    }
+
+    const { data, link } = item;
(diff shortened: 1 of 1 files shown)


packages/core/octo-linker.js
```diff
@@ -18,7 +18,12 @@ function initialize(self) {
 }
 
 function insertDataAttr(matches) {
-  matches.forEach(({ data, link }) => {
+  matches.forEach(item => {
```
stefanbuck: Another solutions would be to strip out all `undefindes` in the first place, although this would require an additional loop over `matches`. Anyway, at the end it really doesn't matter IMO

PR #456, opened 2018-04-01
preventDefault in click handler

xPaw:
Hopefully this fixes scrolling thing popping up when using middle click.

### DIFF
--- a/packages/core/click-handler.js
+++ b/packages/core/click-handler.js
@@ -75,6 +75,8 @@ async function onClick(event) {
     return;
   }
 
+  event.preventDefault();
+
   const $tooltipTarget = $('span', event.currentTarget);
 
   showTooltip($tooltipTarget, PROCESS);


josephfrazier:
Can you elaborate on the issue this fixes?

xPaw:
Should fix the cursor changing to scrolling behaviour when middle clicking. I can't actually test the changed because the extension is refusing to work when I load it locally.

josephfrazier:
Thanks for the explanation (and especially the animation of) the problem! This change seems fine to me, but I'll give it a little more time for @stefanbuck to check out, especially since my machine doesn't show the scrolling behavior when I middle click. You might be able to test your changes in an isolated browser instance by running `yarn chrome-open` or `yarn firefox-open`.

stefanbuck:
@xPaw thanks for the fix, but I would prefer to close this PR for the following reason. Soon there will be a new version out which change the entire current behaviour how linking works in OctoLinker. In fact it will use "real" links which will solve a few core issues along with the cursor bug.

xPaw:
Fair enough.

stefanbuck:
Thanks for your understanding.

josephfrazier:
@stefanbuck, assuming this doesn't break anything, I think we should merge it, despite the fact that a new version will be out soon, since things can always take longer than expected.

What do you think?

stefanbuck:
That's true, it should harm. I still works (there is not difference) on Mac OSX. @xPaw could you please test this as suggested by @josephfrazier with either `yarn chrome-open` or yarn `firefox-open`. Please let us know if you have still issues with testing it locally.

PR #457, opened 2018-04-07
Prevent scroll cursor when using middle mouse click

stefanbuck:
Follow up on #456 

cc @xPaw I've attached a build with the fix applied. Please let me know if this is fixing your issue. Thank you

[octolinker-scroll-fix.zip](https://github.com/OctoLinker/OctoLinker/files/1886286/octolinker-scroll-fix.zip)

### DIFF
--- a/packages/core/click-handler.js
+++ b/packages/core/click-handler.js
@@ -75,6 +75,8 @@ async function onClick(event) {
     return;
   }
 
+  event.preventDefault();
+
   const $tooltipTarget = $('span', event.currentTarget);
 
   showTooltip($tooltipTarget, PROCESS);


xPaw:
> cc @xPaw I've attached a build with the fix applied. Please let me know if this is fixing your issue. Thank you

Welp. It doesn't fix the issue I initially created the PR for...

stefanbuck:
Mhhhh ... I don't have any idea how to fix this, especially because I can't reproduce it. Do you mind to ignore this issue for now? Sorry for the back and forth on this.

xPaw:
Yeah it's not a blocker issue.

PR #463, opened 2018-05-03
Add support for non repo root Haskell projects.

kfigiela:
In my use case we keep several Haskell projects in one repository (monorepo approach). This PR extends search to modules relative to Haskell project root (not git root). It is detected by cutting current path on `src`, `lib`, `app`, or `test`. Also adds navigation in test suites.

### DIFF
--- a/packages/plugin-haskell/__tests__/index.js
+++ b/packages/plugin-haskell/__tests__/index.js
@@ -1,18 +1,34 @@
 import assert from 'assert';
-import githubSearch from '@octolinker/resolver-github-search';
+import { hoogleSearch } from '@octolinker/resolver-hoogle-search';
 import Haskell from '../index';
 
 describe('haskell', () => {
-  const path = '/user/repo/blob/d6/lib/plugins/javascript.js';
-  const target = 'Foo.Bar';
+  it('resolves links - top level project', () => {
+    const path = '/user/repo/blob/v0.1/src/Main.hs';
+    const target = 'Foo.Bar';
 
-  it('resolves links', () => {
     assert.deepEqual(Haskell.resolve(path, [target]), [
-      '{BASE_URL}/user/repo/blob/master/src/Foo/Bar.hs',
-      '{BASE_URL}/user/repo/blob/master/lib/Foo/Bar.hs',
-      '{BASE_URL}/user/repo/blob/master/Foo/Bar.hs',
-      githubSearch({ path, target }).toString(),
-      'https://hackage.haskell.org/package/base/docs/Foo-Bar.html',
+      '{BASE_URL}/user/repo/blob/v0.1/src/Foo/Bar.hs',
+      '{BASE_URL}/user/repo/blob/v0.1/lib/Foo/Bar.hs',
+      '{BASE_URL}/user/repo/blob/v0.1/app/Foo/Bar.hs',
+      '{BASE_URL}/user/repo/blob/v0.1/test/Foo/Bar.hs',
+      '{BASE_URL}/user/repo/blob/v0.1/Foo/Bar.hs',
+      hoogleSearch({ target }).toString(),
+    ]);
+  });
+
+  it('resolves links - monorepo project', () => {
+    const path =
+      '/user/repo/blob/v0.1/projects/project-in-monorepo/src/Main.hs';
+    const target = 'Foo.Bar';
+
+    assert.deepEqual(Haskell.resolve(path, [target]), [
+      '{BASE_URL}/user/repo/blob/v0.1/projects/project-in-monorepo/src/Foo/Bar.hs',
+      '{BASE_URL}/user/repo/blob/v0.1/projects/project-in-monorepo/lib/Foo/Bar.hs',
+      '{BASE_URL}/user/repo/blob/v0.1/projects/project-in-monorepo/app/Foo/Bar.hs',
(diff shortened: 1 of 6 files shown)


packages/plugin-haskell/index.js
```diff
@@ -7,9 +7,17 @@ export default {
   resolve(path, [target]) {
     const [, user, repo] = path.split('/');
     const filePath = target.replace(/\./g, '/');
+    const basePath = path.split(/\/(src|lib|app|test)\//)[0];
+
     return [
+      `{BASE_URL}${basePath}/src/${filePath}.hs`,
+      `{BASE_URL}${basePath}/lib/${filePath}.hs`,
+      `{BASE_URL}${basePath}/app/${filePath}.hs`,
+      `{BASE_URL}${basePath}/test/${filePath}.hs`,
```
stefanbuck: This order is the actual loading order and I wonder if this needs to be moved right before `githubSearch`. OctoLinker does a HTTP Head request for every single url until the first returns `200`. Therefor the most common scenario needs to be checked first.

packages/plugin-haskell/__tests__/index.js
```diff
@@ -8,8 +8,14 @@ describe('haskell', () => {
 
   it('resolves links', () => {
     assert.deepEqual(Haskell.resolve(path, [target]), [
+      '{BASE_URL}/user/repo/blob/d6/src/Foo/Bar.hs',
+      '{BASE_URL}/user/repo/blob/d6/lib/Foo/Bar.hs',
+      '{BASE_URL}/user/repo/blob/d6/app/Foo/Bar.hs',
+      '{BASE_URL}/user/repo/blob/d6/test/Foo/Bar.hs',
```
stefanbuck: `d6` is a branch name, was this intentionally?

kfigiela: I have just replicated existing tests. Will improve.

packages/plugin-haskell/index.js
kfigiela: Actually, scenario I added handles not only monorepo approach, but will also work fine with single project repos. I did some testing and it usually matches at first hit. Bonus point of this is that we browse current branch/commit and don't force master.

stefanbuck: Ahh I see, that sounds great. Can you provide me a link to a mono repo so I can have a look by myself?

stefanbuck:
@kfigiela any updates on this? Would like to release a new version soon and this would be awesome

kfigiela:
Sorry for the delay, got busy with other stuff. Thanks for reminding. Will work on this today!

stefanbuck:
No worries 😉 Please don't forget to post a demo link to a mono repo so I can have a look as well how this will behave. Thank you

kfigiela:
@stefanbuck, pushed new changes and updated tests. 

Summary:
* do only fallbacks on current git revision, this removes unnecessary extra lookups,
* removed GitHub search fallback – it didn't work at all, In fact GH search does not work well for Haskell code and current search API does not allow for reliable queries,
* added fallback to Hoogle search engine (it indexes all packages published on Hackage [Haskell package repository]).

Demo repos: 
* monorepo: https://github.com/haskell-servant/servant – all packages depend on "servant" package, so it's best for testing, note that linking between packages in monorepo will not work (unless GitHub allows for path suffix queries).
* single package repo: https://github.com/TaktInc/freer/ – most of modules in this package are in `Eff.` namespace

stefanbuck:
@kfigiela looks good so far. I added three commits on top to change a few minor details. Thank you for all your effort and thinking on this PR 👍

packages/resolver-hoogle-search/index.js
```diff
@@ -18,11 +18,10 @@ export function handleResponse(hoogleSearchUrl, target, response) {
 
 export function hoogleSearch({ target }) {
   const query = `${encodeURIComponent(target)}%20is:module%20is:exact`;
-  const hoogleSearchUrl = `https://hoogle.haskell.org/?hoogle=${query}`;
-  const apiUrl = `${hoogleSearchUrl}&mode=json`;
+  const hoogleSearchUrl = `https://hoogle.haskell.org/?hoogle=${query}&mode=json`;
```
kfigiela: This simplification is not correct. Note that we use `apiUrl` for JSON API lookup. However, in case that more than one package provides module with the same name I redirect to search page that indicates matches (`hoogleSearchUrl` that lacks `&mode=json` parameter). Now it redirects to JSON response.

Example use case: follow `Net.IPv4` import on https://github.com/rlupton20/vanguard-dataplane/blob/293e2e34c31b17051baa5cf02824c275aa82e772/app/Debug/PacketParsing/IP4.hs
Now it redirects to https://hoogle.haskell.org/?hoogle=Net.IPv4%20is:module%20is:exact&mode=json which is not very useful, however search page looks better https://hoogle.haskell.org/?hoogle=Net.IPv4%20is:module%20is:exact

kfigiela:
@stefanbuck Thanks for releasing this so quickly! I noticed that API url simplification introduced a small bug. See my comment in the commit https://github.com/OctoLinker/OctoLinker/pull/463/commits/cc7fc70f55b40522432af6e594bcf239ca8e7fd7#r192562811

stefanbuck:
Oh nooo, sorry about that! I created a PR to fix it #470

kfigiela:
Thanks for fixing this!

PR #465, opened 2018-05-14
Improve Python support

stefanbuck:
Today I noticed that our Python plugin does not support the following import style:

```py
from object_detection.utils import label_map_util
```
_[Source](https://github.com/tensorflow/models/blob/master/research/object_detection/eval.py#L56)_

This should resolve to https://github.com/tensorflow/models/blob/master/research/object_detection/utils/label_map_util.py.

To achieve this the Python RegExp needs to be modified to take the second value into account. I updated the test, but I struggle with updating the related RegExp. @josephfrazier are you able to help with this RegExp?


josephfrazier:
I updated the RegExp so that the second value matches in most of the test cases, but it looks like the example link still doesn't resolve, so I guess there's more to be done in the resolution logic: https://github.com/tensorflow/models/blob/9f58547c45057d0d1e0848cb0d53c2b516d8784f/research/object_detection/eval.py#L56

stefanbuck:
Thanks for completing the RegExp @josephfrazier. Yes, there is a little bit more effort needed to make this working. The related resolver needs to evaluate the second value as well, which shouldn't be that hard to achieve I think . Will continue working on this next week.

stefanbuck:
The example link from above does still not work, but I added a few other cases which didn't worked before:

https://github.com/keras-team/keras/blob/2eed2a5d688e8720740f30d281a18fb2c13a15fd/keras/__init__.py#L22
https://github.com/keras-team/keras/blob/2eed2a5d688e8720740f30d281a18fb2c13a15fd/keras/models.py#L8

josephfrazier:
Nice! Should we go ahead and merge this, then open another issue for the original link?

stefanbuck:
Yes, good idea!

PR #466, opened 2018-05-23
Fix openTab handler for Firefox

victorb:
Before this commit, right-clicking (or middle-mouse clicking) on a link
doesn't open it in a new tab in Firefox. This was because the extension
is using a chrome-specific module for the storage.

Instead, I've changed it to use the webextension-polyfill, which makes
the storage APIs work the same in Chrome and Firefox, so you don't have
to care about platform-specific APIs.

License: MIT
Signed-off-by: Victor Bjelkholm <git@victor.earth>

### DIFF
--- a/packages/helper-settings/index.js
+++ b/packages/helper-settings/index.js
@@ -1,4 +1,4 @@
-/* global browser */
+import browser from 'webextension-polyfill';
 import uuid from 'uuid';
 
 const store = {};
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -17,7 +17,6 @@
   },
   "background": {
     "scripts": [
-      "node_modules/webextension-polyfill/dist/browser-polyfill.js",
       "background.js"
     ]
   },


packages/helper-settings/index.js
```diff
@@ -1,7 +1,6 @@
+/* global browser */
```
stefanbuck: Have you tried to import the polyfill right here? I think this should work and so it will be part of the webpack build.

VictorBjelkholm: I don’t not try that, just followed the instructions from the readme. But I’ll give it a try and report back, thanks!

josephfrazier: Yes, I think you'll need to import it here, (instead of referencing it inside `node_modules` [above](https://github.com/OctoLinker/OctoLinker/pull/466/files#diff-76e85634924729360ea8e5952981214dR20)) to make sure it's present in the build delivered to end users (rather than just the development version).

PR #467, opened 2018-05-31
Fix bower, composer, and dotnet plugins

mrliptontea:
Hi,

I noticed that OctoLinker is throwing an error when viewing composer.json and doesn't work in general with it. Short inspection showed that some of the changes made in March were not reflected in plugins using @octolinker/helper-process-json which lead to finding few more issues, but in the end and got it to work.

Let me know if anything needs improving.

### DIFF
--- a/packages/core/octo-linker.js
+++ b/packages/core/octo-linker.js
@@ -42,6 +42,8 @@ function run(self) {
     });
   });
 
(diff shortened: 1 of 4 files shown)


stefanbuck:
How could I miss that? Thank you so much for . Will release a new version either later today or tomorrow.

PR #472, opened 2018-06-16
Speed up linking 

stefanbuck:
Yesterday, I noticed that in some cases, I have to wait quite long time for a redirect. For example ur [JavaScript resolvers](https://github.com/OctoLinker/OctoLinker/blob/master/packages/plugin-javascript/index.js#L21-L59) returns an Array of possible urls which is sort by the most common cases. 

```js
[
    '{BASE_URL}path/file.js',
    '{BASE_URL}path/file/index.js',
    '{BASE_URL}path/file.jsx',
    '{BASE_URL}path/file/index.jsx',
    '{BASE_URL}path/file.ts',
    '{BASE_URL}path/file/index.ts',
    '{BASE_URL}path/file.tsx',
    '{BASE_URL}path/file/index.tsx',
    '{BASE_URL}path/file.ls',
    '{BASE_URL}path/file/index.ls',
    '{BASE_URL}path/file.json',
    '{BASE_URL}path/file/index.json',
    '{BASE_URL}path/file',
]
```

This makes resolving less popular cases quite slow. In [our repository](https://github.com/OctoLinker/OctoLinker/blob/master/packages/core/notification.js#L4) we have `require('../package.json')`. A click on this will load 10 possible urls before it hits the actual loadable path.

To solve this issue, I added a new helper package which allows us to sort urls returned by a resolver by file extensions. I couldn't think of any edge cases and therefore I added this new behaviour to core which applies it to all urls by default.

### DIFF
--- a/packages/core/octo-linker.js
+++ b/packages/core/octo-linker.js
@@ -2,6 +2,7 @@ import injection from 'github-injection';
 import BlobReader from '@octolinker/blob-reader';
 import insertLink from '@octolinker/helper-insert-link';
 import * as storage from '@octolinker/helper-settings';
+import helperSortUrls from '@octolinker/helper-sort-urls';
 import notification from './notification';
 import clickHandler from './click-handler';
 import Plugins from './plugin-manager.js';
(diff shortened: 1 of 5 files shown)


josephfrazier:
Wow, nice fix! I had also noticed the delay, but didn't think to look into it. Do you know if there's an easy way to test this? No worries if there's not, it's a huge improvement in performance. Feel free to merge :)

stefanbuck:
Unfortunately not. Integration testing would be really awesome to have, but I didn't had time to look deeper into this.

PR #476, opened 2018-06-17
Adjust resolver to return a single string or an array of string whenever possible

stefanbuck:
A resolver can return either a string, an array of strings, an object or a function. This is quite confusing. This PR aims to  simplify this. Apart from the `github-search` resolver all other resolvers are returning either a string or an array of strings.

### DIFF
--- a/packages/core/click-handler.js
+++ b/packages/core/click-handler.js
@@ -37,25 +37,29 @@ function getResolverUrls(urls) {
       return null;
     }
 
+    // github-search resolver returns a function
     if (typeof url === 'function') {
       return {
         func: url,
       };
     }
 
-    if (typeof url === 'object') {
+    // Live-query resolver results
+    if (url.startsWith('https://githublinker.herokuapp.com')) {
       return {
-        url: url.url.replace('{BASE_URL}', BASE_URL),
-        method: url.method,
+        url,
+        method: 'GET',
       };
     }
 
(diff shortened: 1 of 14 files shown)


packages/core/click-handler.js
```diff
@@ -37,25 +37,29 @@ function getResolverUrls(urls) {
...
       return {
         func: url,
       };
     }
 
-    if (typeof url === 'object') {
+    // Live-query resolver results
+    if (url.startsWith('https://githublinker.herokuapp.com')) {
       return {
-        url: url.url.replace('{BASE_URL}', BASE_URL),
```
josephfrazier: It looks like this change is technically unrelated, but should have been done earlier anyway?

packages/core/click-handler.js
```diff
@@ -59,7 +59,7 @@ function getResolverUrls(urls) {
       };
     }
 
-    // Extneral urls
+    // External urls
```
stefanbuck: 😆 Today, I installed a spell checker plugin! From now on, you won't see that kind of typos 🤞

packages/core/click-handler.js
stefanbuck: Exactly, only the live-query resolver was retuning an object wich didn't have a `{BASE_URL}` placeholder at all.

Issue #277, opened 2017-02-05
Show status bar


tim-phillips:
Chrome Version 56.0.2924.87 (64-bit)
OctoLinker latest

OctoLinker should show browser status bar when mousing over a link

stefanbuck:
That's not possible, unfortunately. We don't know the target URL while rendering the links. The process to find the target URL is in many cases request heavy and therefore we can't do this for every link by default on extension start. The actual click on a link will kick off the process.

stefanbuck:
@babenzele is this a big issue for you and if so why? Is this a trust thing?

babenzele:
@stefanbuck Gotcha re: lazy loading the resource. 

It's not a trust thing. I was curious about where some imports were coming from and thought it could be handy to hover over the import to see the url, mostly to see the github org/user the repo is located, without having to click through.

josephfrazier:
Hmm, it might not be hard to add a `mouseover` listener that starts the URL resolution process and shows the results in a tooltip (which there is already code for).

@babenzele, if you're curious, there's some more discussion of making the links into actual `<a href>` elements in https://github.com/OctoLinker/browser-extension/pull/87 and https://github.com/OctoLinker/browser-extension/issues/49#issuecomment-137948995

derimagia:
Just to throw my opinion on it, this is a pretty major UX issue in my opinion. There's at least https://fiatjaf.alhur.es/module-linker/ which seems to improve this quite a lot, but it seems silly to have 2 plugins doing the same thing. 

I can't say I know much about how octolinker works so I can't comment too much on this besides what i have at the moment.

stefanbuck:
I discovered module-linker a few weeks ago and was looking how they do linking. As assumed it prefetches all links on startup which is the major difference between OctoLinker and module-linker. 

@josephfrazier suggested the same technic a while ago, but I wasn't confidence. Mainly because I was afraid to put that many unnecessary traffic on our service, github and all the registries. In worse case one of them would ban our service. Maybe worth to revisit and investigate into it. 

[...]

PR #477, opened 2018-06-17
Add integration tests

stefanbuck:
This is a first implementation to load the real OctoLinker browser extension in a Chrome instance using puppeteer.

I've created a separate repository which contains all the test files https://github.com/OctoLinker/e2e. End to end tests are generated on-the-fly based on the repo content so no additional PR in here is required to increase our coverage. 

Before `yarn e2e` invokes the actual tests, it downloads https://github.com/OctoLinker/e2e/blob/master/resource.json which is used to generate our e2e tests. For more details visit https://github.com/OctoLinker/e2e


jest-puppeteer.config.js
```diff
@@ -0,0 +1,15 @@
+const path = require('path');
+
+const extPath = path.join(__dirname, './dist');
+
+module.exports = {
+  launch: {
+    dumpio: true,
+    headless: false,
```
stefanbuck: Puppeteer headless doesn't support extensions, right now.
 
 https://github.com/GoogleChrome/puppeteer/issues/2819#issuecomment-401432184

stefanbuck:
Hi @josephfrazier 👋 Friendly reminder in case you missed my review request. I'm excited to hear feedback on this

josephfrazier:
Thanks for the reminder, I had forgotten about this one. This looks really good overall! I'm excited to see proper Chrome tests, and I've been looking for a reason to get familiar with `puppeteer`.

I do have one question though: Why put the e2e tests in a separate repo? It seems like that makes it harder to keep track of them, versus just keeping them here.

stefanbuck:
No worries, happens to me all the time 😉 Puppeteer is really amazing and defiantly something you should look into. But back to topic: 

That's a good question. Initially I thought it would be good just for the sake of separating. Also I was thinking about cases were we have to replicate some special file tree structure. However, right now there is no need for this  and maybe never will be. The only last thing I can think of is when it comes to test functionality related to issues or PullRequests. Such dummy issues / PR I would see in this repo.

Conclusion, I'll merge https://github.com/OctoLinker/e2e into this and we will use the `e2e` repo to test other GitHub related features.

stefanbuck:
I deleted that E2E repository and merged all the files into `e2e/fixtures/`. On a PullRequest Puppeteer runs the test against the fork so we can verify that fixtures files are working before we merge the PR. I also added a README to explain how it works. Please free to go wild with commenting typos and other suggestions you have to make this useful for others.

stefanbuck:
Yes, I really put a lot of thinking into it, while I had a couple days off 😉 What would you snapshots then? We don't store any information in the DOM, so I don't know what you want to snapshots. 

Once #277 is resolved, we can think about snapshotting the DOM. Although we have to be really carefully what we snapshot to keep those small as possible.

stefanbuck:
@josephfrazier I'm going to merge this PR now. Let's keep eye on snapshot testing once #277 is complete. Are you able to review the copy in e2e/readme.md raise a fix PR if needed? That would be really awesome.

josephfrazier:
Sounds good, sorry I didn't respond earlier about the snapshotting. I was thinking we can use snapshots to store the expected test results, instead of using the `"//": "..."` "comments". Jest snapshots aren't restricted to (V)DOM objects, so it should be doable. Maybe I'll try to set it up if I find the time.

I'll take a look at the E2E readme now and suggest changes.

PR #478, opened 2018-06-28
go: add k8s.io support

evanlucas:
This adds support for go imports that use k8s.io.

https://github.com/kubernetes/ingress-gce/blob/master/cmd/glbc/main.go can be used to see the before/after changes

Thanks for an awesome extension!

### DIFF
--- a/packages/helper-grammar-regex-collection/go.js
+++ b/packages/helper-grammar-regex-collection/go.js
@@ -9,6 +9,7 @@ const ALLOWED_DOMAINS = [
   'hub.jazz.net',
   'gopkg.in',
   'golang.org',
+  'k8s.io',
 ];
 
 function multiImportRegExpBuilder(input) {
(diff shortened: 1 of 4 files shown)


stefanbuck:
It's supposed to be supported and I'm happy to tell you that this is just a bug in our service which is doing the linking. Will be fixed shortly. https://github.com/OctoLinker/live-resolver/pull/71

packages/helper-grammar-regex-collection/go.js
```diff
@@ -9,6 +9,7 @@ const ALLOWED_DOMAINS = [
   'hub.jazz.net',
   'gopkg.in',
   'golang.org',
+  'k8s.io',
```
stefanbuck: Darn, I thought `k8s.io` is already part of this list. Will re-open your PR. This should be the only change which is needed to add k8s.io support since all the linking is made through https://github.com/OctoLinker/live-resolver/

stefanbuck:
Hey there 👋, are you still planing to contribute this features @evanlucas? Let me know if you need help.

evanlucas:
Yes! Sorry I’m on vacation until next week though! If you want to land before then, feel free to fix it up however it needs. I’m not worried about the contribution. Otherwise, I should be able to get this updated on Monday.

stefanbuck:
Closing this in favour of #480

stefanbuck:
Hi @evanlucas, it took me a bit longer than expected to release this feature. A new version 4.19.0 will be available in a few minutes.

PR #479, opened 2018-07-04
Add helper-array-reorder

stefanbuck:
That's a tough one and looks weird, but trust me it makes sense. 

Since a few moths already I'm doing little refactoring here and there downwards to solve #277. This helper package is not in use yet, but in order to make PRs easier to review I'll break down all the work that I did in https://github.com/stefanbuck/OctoLinker/tree/277-real-links into small chunks. 

Back to the this PullRequest. As the readme says, it's about reordering an array based on the indexes defined in another array. I think it's best to take a look at the example or tests. 

### Why do we need this?

When we resolve all url after page load, we have to do many requests in order to figure the locations for relative files. I described the reasons for this in this comment https://github.com/OctoLinker/OctoLinker/issues/277#issuecomment-301438244 

However, this adds some kind learning algorithm while resoling urls. Whenever a possible url performs a successful HTTP HEAD request we increase that array index and store the count in another array.

Given we have an import like `require('./lib/loader');`. The possible urls array for this looks like this:

```js
[
  './lib/loader.js'        // does not match
  './lib/loader/index.js'  // does not match
  './lib/loader.ts'        // does not match
  './lib/loader/index.ts'  // Match
]
```

This will push 3 into an array and before we run the next resolver this array will be sorted again which reduces the request for some cases a lot. 

```js
[
  './lib/loader/index.ts'   
  './lib/loader.js'       
  './lib/loader/index.js' 
  './lib/loader.ts' 
]
```

This assumes that a import style within an repo is consistent and the majority wins the race.

### DIFF
--- a/packages/helper-array-reorder/README.md
+++ b/packages/helper-array-reorder/README.md
@@ -1,0 +1,15 @@
+# @octolinker/helper-array-reorder 
+
+This array helper sorts an array by passing another array which define the new order by the array index of the items you want to sort and retains all unmatched array indexes.
+
+
+```js
+// Push Fri the to beginning, because the sort array starts with the array index of 4 which is Fri
+sortResolveUrls(['Mo', 'Tue', 'Wed', 'Thu', 'Fri'], [4]); // ['Fri', 'Mo', 'Tue', 'Wed', 'Thu']
+
+// Push Fri the to beginning followed by Tue
+sortResolveUrls(['Mo', 'Tue', 'Wed', 'Thu', 'Fri'], [4, 1]); // ['Fri', 'Tue', 'Mo', 'Wed', 'Thu']
+
+// Let's revers all weekdays 
+sortResolveUrls(['Mo', 'Tue', 'Wed', 'Thu', 'Fri'], [4, 3, 2, 1, 0]); // ['Fri', 'Thu', 'Wed', 'Tue', 'Mo']
+```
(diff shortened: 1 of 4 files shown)


josephfrazier:
Hey, this looks like an interesting approach to handling module resolution! When this package is actually used, will there by a separate array of URLs stored for every path/link combination? I'd like to make sure I understand how things are going to work.

stefanbuck:
No, not really. As you know, a resolver returns an array of possible urls. For JavaScript this array is quite big and for others it's maybe just one or two urls. However, the stored index matches the position of an url within this array. If you have three imports and the first two are the same (resolver url wise), it will also load the third url with `../index.js` before it tries the other urls. 

```js
import "./a"  // ./a/index.js
import "./b"  // ./b/index.js
import "./c"  // ./c.js
```

Does this answer your questions? There is a wip example in the branch that I linked above https://github.com/stefanbuck/OctoLinker/commit/638b07b9c63a16d75b72eab64ac93aeb9ec9f582#diff-02427ff47ccc6080732a1691b32ffaf0R39

stefanbuck:
It may has side effects when the same resolver returns different arrays based on the input which is the case for JavaScript. Was your question referring to this scenario?

stefanbuck:
Closing this for now, since I need time to think how to solve the mentioned issue above.

PR #480, opened 2018-07-07
Add support for k8s.io

stefanbuck:
Replacement for #478 to add support for kubernetes imports. 

Test url: https://github.com/kubernetes/ingress-gce/blob/6f8bdc9fad1b5576d4fe07900301a133652a8892/cmd/glbc/main.go#L31

### DIFF
--- a/packages/plugin-go/index.js
+++ b/packages/plugin-go/index.js
@@ -1,5 +1,6 @@
 import { join, dirname } from 'path';
 import { go } from '@octolinker/helper-grammar-regex-collection';
+import liveResolverQuery from '@octolinker/resolver-live-query';
 
 function goFile({ path, target }) {
   const list = [];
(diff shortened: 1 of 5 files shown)


packages/plugin-go/index.js
```diff
@@ -47,7 +48,11 @@ export default {
       return githubUrls(target);
     }
 
-    return [`https://${target}`, `https://golang.org/pkg/${target}`];
+    return [
+      `https://${target}`,
+      `https://golang.org/pkg/${target}`,
+      liveResolverQuery({ type: 'go', target }),
```
stefanbuck: I thought we're already using https://github.com/OctoLinker/live-resolver for `go` packages as a fallback 😄

stefanbuck:
Another simple change to add support for kubernetes imports. Merge this by myself too

josephfrazier:
:+1:

PR #481, opened 2018-07-07
Remove client side tracking

stefanbuck:
A few years ago, I added client side tracking to get more insights about the usage and potential issues with OctoLinker. The value of those reports is quite low and I didn't check those on a regular basis. It's time to remove client side tracking to satisfy our users which care about their privacy.


stefanbuck:
It's a trivial change so I'm gonna to merge it by myself

josephfrazier:
:+1:

PR #484, opened 2018-07-18
Update E2E readme grammar

josephfrazier:
As requested in https://github.com/OctoLinker/OctoLinker/pull/477#issuecomment-406057799

I also think we should change the "How do we know" heading, but I'm not sure what to change it to. @stefanbuck, could you explain the concept you're trying to write? Thanks!


josephfrazier:
Thanks!

PR #488, opened 2018-07-30
Link npm dependecies in Pull Request

### LINKED ISSUE #71: Npm dependencies from package.json do not work in Pull requests
npm deps work in file view

image

but don't it PR view

image

btw thanks for awesome work

stefanbuck:
This turned into a slightly bigger Pull Request than I usually prefer. In order to support `package.json` `bower.json` and `composer.json` files in a Pull Request diff view, I had to make changes all over the place. I'll quickly outline what I did and why this was necessary.

### Why this was not working

Our plugins for the mentions files above relay on a well formatted json object. This is the case when looking at a single file, but not in a Pull Request view which is in most cases just a subset of the actual file. 

### How I solved it

In order to fix #71 I had to ensure that we always have a valid JSON file to operate on. Therefore I added two methods to our [blob-reader](https://github.com/OctoLinker/OctoLinker/tree/master/packages/blob-reader) package. 

#### `fetchBlob`
This method allows us to fetch the raw underlay blob of the current blob. We only do this on a Pull Request view.

#### `fetchParentBlob`
Along with this method, I added a new concept of parent blobs were a blob can reference to another parent blob. We need this in order to link all values on the left and also on the right side of the diff view. 

Those two methods are called right before we invoke a plugin and as I already mentioned only on a diff view. Apart from that, I added a few missing tests.

### How to review this Pull Request best?
I would suggest to go through commit by commit.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 dist/
 packages/**/node_modules
 e2e/fixtures/
+coverage/
--- a/_jestsetup.js
+++ b/_jestsetup.js
@@ -1,7 +1,10 @@
 import fs from 'fs';
 import path from 'path';
+import fetchMock from 'jest-fetch-mock';
 
 jest.mock('path');
+
+global.fetch = fetchMock;
 
 global.chrome = {};
 global.chrome.runtime = {
--- a/e2e/automated.test.js
+++ b/e2e/automated.test.js
@@ -1,8 +1,7 @@
 const fixtures = require('./fixtures.json'); // eslint-disable-line import/no-unresolved
+const diffFixtures = require('./diff-fixtures.json'); // eslint-disable-line import/no-unresolved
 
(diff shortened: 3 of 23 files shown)


stefanbuck:
@josephfrazier this is ready to be looked at. Since we don't have E2E tests for the diff view, it would be great if you could also do a quick sanity check on this. Thank you.

stefanbuck:
@josephfrazier I know you're super busy and I therefore I wonder if you are happy with me merging this without your review? Would still appreciate your input on it one day 😉 Before I merge this, I'll do another QA round by myself.

josephfrazier:
Yeah, sorry, I've been busy with work and life recently. As long as it passes all the pre-existing tests, go for it! :)

stefanbuck:
No worries, no problem at all, @josephfrazie.

PR #490, opened 2018-07-31
Add Codecov

stefanbuck:
Note: I branched off from #489 to get a working build

Configure this project to report code coverage using codecov https://codecov.io/gh/OctoLinker/OctoLinker

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ lerna-debug.log
 dist/
 out/
 e2e/fixtures.json
+coverage/
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_script:
   - ./packages/blob-reader/scripts/update-fixtures.sh
   - yarn package
 script:
-  - travis_retry yarn test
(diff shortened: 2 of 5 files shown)


josephfrazier:
Hmm, it can't hurt, right? Is it configured to require PRs to include coverage?

stefanbuck:
No, it shouldn't hurt. 

> Is it configured to require PRs to include coverage?

No, I don't think so. I didn't set up anything in this direction. Btw. are your sure that codecov supports this or was it just a general question?

josephfrazier:
I'm not completely sure, but I think it's possible to require PRs to not decrease coverage. I doubt that's the default though, so we should be good.

stefanbuck:
I quickly scanned their docs, but I couldn't find anything. https://docs.codecov.io/docs/codecov-yaml

Anyway, let's see if this works for us. We can easily remove it anytime. Thanks for your feedback

josephfrazier:
Of course, I'm looking forward to seeing how it goes!

Issue #128, opened 2016-07-16
Firefox extension is not automatically published


josephfrazier:
https://github.com/OctoLinker/browser-extension/pull/119 added automatic publishing to the Chrome extension store, but we don't yet have the same for Firefox. Something like https://github.com/kewisch/pyamo might be able to help with this.

EDIT: I also found https://www.npmjs.com/package/firefox-extension-deploy

prometh07:
Maybe https://github.com/mozilla/web-ext ?

bfred-it:
web-ext can build, lint, run and sign, but not deploy for now: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference

josephfrazier:
It looks like @bfred-it ^ has found a way to do this with `web-ext`: https://github.com/mozilla/web-ext/issues/804#issuecomment-302588357

PR #493, opened 2018-08-10
Setup AMO auto deploy

stefanbuck:
Hi @bfred-it, since you already have quite experience with setting up AMO deploy, do you mind to review this Pull Request? That would be really awesome.

Do I have to setup any environment variables?

Fix for #128


package.json
```diff
@@ -13,7 +13,9 @@
     "test:watch": "jest --watch",
     "version": "json -I -f assets/manifest.json -e \"this.version='`json -f package.json version`'\" && git add assets/manifest.json",
     "package": "yarn build && web-ext build --source-dir dist --artifacts-dir out --overwrite-dest",
-    "release": "webstore upload --source dist --auto-publish",
+    "release:cws": "webstore upload --source dist --auto-publish",
```
bfred-it: I'd probable use `cd dist && webstore upload --auto-publish` just to match the line at the bottom and avoid a flag.

stefanbuck: Good point :+1:

PR #498, opened 2018-08-21
Add new issue templates

stefanbuck:
Somehow I missed this new feature before and saw it today for the first time. 

https://blog.github.com/2018-05-02-issue-template-improvements/


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,25 @@
...
+name: 🐛 Reporting a Bug
+about: Open a new issue here if something isn't working as expected.
+---
+
+<!--
+  Thanks for filing an issue!
+
+  Please make sure that you include the following information to ensure that your issue is actionable.
+-->
+
```
josephfrazier: I think we should probably include the following fields from the old issue template here:

```markdown
**Browser name:** 	

**Browser version:** 	

**OctoLinker version:** 	

**URL and line number where issue occurs:**
```

josephfrazier: I think we should probably include the following fields from the old issue template here:

```markdown
**Browser name:**

**Browser version:**

**OctoLinker version:**

**URL and line number where issue occurs:**
```

stefanbuck: Really? I never found them useful and I think it's annoying to fill them out too. In 99% the OctoLinker version is equal to the latest available version (thanks auto update). Apart from one or two issues we never had issues with a specific browser in a particular version. Most of the times bug reports are related to missing features or unexpected behaviour. However, if you have strong feelings about adding browser version and name, I'm happy to add them back.

josephfrazier: Hmm, good points. Perhaps I wasn't remembering correctly. I see that we already have the URL/line mentioned below, so this should be ok!

stefanbuck:
Great, thanks for your feedback anyway.

PR #499, opened 2018-08-22
Support JS import in Vue files

leomoty:
This implements https://github.com/OctoLinker/OctoLinker/issues/496, which allows to jump to .js files from .vue.

This link should work:
https://github.com/vuejs/vue-hackernews-2.0/blob/master/src/components/Item.vue#L30

I'm not entirely sure regarding tests, can you point me how that should work?

### Checklist:

### DIFF
--- a/packages/plugin-javascript/index.js
+++ b/packages/plugin-javascript/index.js
@@ -92,6 +92,7 @@ export default {
         /\.es6$/,
         // CoffeeScript
         /\.coffee$/,
+        /\.vue$/,
       ],
       githubClasses: [
         'type-javascript',
@@ -100,6 +101,7 @@ export default {
         // CoffeeScript
         'type-coffeescript',
         'highlight-source-coffee',
+        'type-vue',
       ],
     };
   },


packages/plugin-javascript/index.js
```diff
@@ -20,7 +20,7 @@ function getTopModuleName(target) {
 
 export function javascriptFile({ path, target }) {
   const list = [];
-  const extName = ['.js', '.jsx', '.ts', '.tsx', '.ls', '.json'];
+  const extName = ['.js', '.jsx', '.ts', '.tsx', '.ls', '.json', '.vue'];
```
stefanbuck: Adding `.vue` here is not need unless you want to support such a import `import item.vue`. I never worked with Vue so not sure if this is a common case or not.

packages/plugin-javascript/index.js
```diff
@@ -92,6 +92,7 @@ export default {
         /\.es6$/,
         // CoffeeScript
         /\.coffee$/,
+        /\.vue?$/,
```
stefanbuck: The file extensions is always `.vue`, right? If so, you don't need the question mark otherwise the character in front of it will become optional.

packages/plugin-javascript/index.js
stefanbuck: Reverting that change also makes the tests passing, since the other change is not covered by any tests which isn't a big deal in my opinion.

leomoty: Right, I'll update that, I kinda misunderstood your comment at the issue.

packages/plugin-javascript/index.js
leomoty: Will update.

packages/plugin-javascript/index.js
stefanbuck: My fault, I thought that's needed to support Vue.

leomoty:
Updated & force pushed.

PR #500, opened 2018-08-22
Update puppeteer related dependecies to lastest version

stefanbuck:
We receive a lot timeouts from our E2E tests recently. I already increased the timeout to 20 sec, but they are still failing. Updating puppeteer and all related dependecies like `jest-puppeteer` and `jest` seems to fix this issue


stefanbuck:
Ohh no the second CI run was failing. Anyway, upgrading those dep. is still  useful

PR #508, opened 2018-08-29
Fix yarn chrome-open to run in Windows

### LINKED ISSUE #507: yarn *-open commands
**Intended outcome:**

Running `yarn chrome-open` should open a chrome session with the extension loaded

**Actual outcome:**

Complains about an issue with the manifest

**How to reproduce the issue:**

Run `yarn chrome-open`

----

`yarn firefox-open` on other hand seems to work fine, but OctoLinker is not injected properly in some pages, like the diff view.

leomoty:
Fixes #507, unsure if you want to keep these *-open around, but in Windows the relative dir is not resolved.

### Checklist:

### DIFF
--- a/scripts/chrome-launch.js
+++ b/scripts/chrome-launch.js
@@ -1,9 +1,11 @@
 #!/usr/bin/env node
 
 const chromeLaunch = require('chrome-launch'); // eslint-disable-line import/no-extraneous-dependencies
+const path = require('path');
 
 const url = 'https://github.com/OctoLinker/OctoLinker/blob/master/package.json';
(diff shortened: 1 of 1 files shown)


stefanbuck:
Ohh Travis is failing, because GitHub now lazy loads outdated comments. 



Maybe they just evaluating a new approach and I think we should wait until we tackle this. Therefore I think it's best to skip this failing for now.

stefanbuck:
@leomoty please rebase so the failing test gets ignored

josephfrazier:
Thanks for figuring this out, @leomoty!

PR #514, opened 2018-09-14
Little refactoring to avoid circular execution when listing for dom mutations

stefanbuck:
For some reason the following PR https://github.com/microdevs/missy/pull/39/file causes the browser hang when trying to add a inline comment. I tested the same functionally on a few other PRs and it worked fine for all of them. However, I was able to narrow it down to the `MutationObserver` which is called over and over again. I have no idea what is different there, but I decided to rewrite the related code to only spy for dom mutations when expanding the blob.


stefanbuck:
Sure I did and it's also covered by E2E tests 😃

leomoty:
Nice!

PR #516, opened 2018-09-27
Do not capture whitespace in Dockerfile image link

### LINKED ISSUE #515: Whitespace in dockerfile image links should be ignored
**Intended outcome:**

Clicking on a `FROM image` link with whitespace between the FROM and the image in a Dockerfile and getting redirected to the correct docker hub link.

**Actual outcome:**

There was whitespace between the FROM and the image name, thus I was redirected to an url like `https://hub.docker.com/_/%20%20%20%20%20%20%20ubuntu/` where %20 stands for whitespace (a space).

**How to reproduce the issue:**

Click on the FROM in the [ffmpeg 4.0 dockerfile](https://github.com/jrottenberg/ffmpeg/blob/master/docker-images/4.0/ubuntu/Dockerfile#L8)

selection_999 109

MatthewDG:
### Checklist:

Fixes #515 - the Docker image link should not include the whitespace preceding `ubuntu:16.04 AS base` in https://github.com/jrottenberg/ffmpeg/blob/master/docker-images/4.0/ubuntu/Dockerfile#L8

### DIFF
--- a/e2e/fixtures/docker/Dockerfile
+++ b/e2e/fixtures/docker/Dockerfile
@@ -1,8 +1,5 @@
 // @OctoLinkerResolve(https://hub.docker.com/_/php/)
 FROM php:php:5.6-apache
 
-// @OctoLinkerResolve(https://hub.docker.com/_/ubuntu/)
-FROM        ubuntu:16.04 AS base
-
 // @OctoLinkerResolve(<root>/docker/docker-entrypoint.sh)
 ENTRYPOINT ["docker-entrypoint.sh"]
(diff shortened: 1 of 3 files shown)


e2e/fixtures/docker/Dockerfile
```diff
@@ -1,5 +1,8 @@
 // @OctoLinkerResolve(https://hub.docker.com/_/php/)
 FROM php:php:5.6-apache
 
+// @OctoLinkerResolve(https://hub.docker.com/_/ubuntu/)
+FROM        ubuntu:16.04 AS base
```
stefanbuck: Even I like to see more and more E2E tests, I think we should not test everything end to end. Mainly because they are super slow. 

A rule of thumb could be, one E2E test per feature and unit tests for variations like this. 

rule of thumb would be interested in your thoughts on this?

MatthewDG: I agree, the E2E tests take a long time to run already so it makes sense to not test everything especially if we start to add fixtures for the other plugins.

Issue #510, opened 2018-08-31
Multiline import in diff view


leomoty:
**Intended outcome:**

`dialogx.jsx` is clickable

**Actual outcome:**

`dialogx.jsx` is not clickable in diff view

**How to reproduce the issue:**

https://github.com/cockpit-project/cockpit/pull/9943/files#diff-a2edaced83836966f97b3d8df14beabcR24

Doesn't work on chrome (master) for me, works fine in Firefox.

It works correctly in blob (both browsers):

https://github.com/cockpit-project/cockpit/blob/7f47354e306dc5be08d235be9e3ae52c1f85d445/pkg/storaged/content-views.jsx#L25

stefanbuck:
Good spot 👍@leomoty do you want to look into this one?

A good starting point is to verify that the RegExp search made by [findAndReplaceDOMText](https://github.com/padolsey/findAndReplaceDOMText) is working. The [replace](https://github.com/OctoLinker/OctoLinker/blob/3da6100e2b121a641cd3a7abf900d091f5dd3ad3/packages/helper-insert-link/index.js#L166) callback should be called with this `match` otherwise our advanced [replace handler*](https://github.com/OctoLinker/OctoLinker/blob/3da6100e2b121a641cd3a7abf900d091f5dd3ad3/packages/helper-insert-link/index.js#L92) can't do his job.

\* Warning in advance: The advanced replace handler is the most complicated part in OctoLinker 😉

leomoty:
Sure, will take a look this weekend. Any ideas why it works on firefox but not chrome?

leomoty:
So, I was able to narrow down the issue, it only fails on split view:

https://github.com/cockpit-project/cockpit/pull/9943/files?utf8=%E2%9C%93&diff=split

It happens in both chrome and firefox.

leomoty:
@stefanbuck  I had a hunch.

When in split view, this is the HTML structure is as follow:

```
<tr>
    <td class="blob-num blob-num-deletion js-linkable-line-number" id="diff-a2edaced83836966f97b3d8df14beabcL24" data-line-number="24"></td>

    <td class="code-review blob-code blob-code-deletion ">

[...]

PR #517, opened 2018-10-02
Rerun octolinker after progressive-container has loaded

### LINKED ISSUE #449: Not all supported NuGet references are being linkified
**Browser name:** Chrome

**Browser version:** 64

**OctoLinker version:** 4.16.1

**URL and line number where issue occurs:** https://github.com/equineexchange/EquineExchange.ImageHosting/commit/3a917291b3e2ad459d2ed84af96b6245658758bf#diff-f2c1409a20722ddaf5705add38f2fbe1

**Expected behavior:** The NuGet packages should be linkified like they are elsewhere such as https://github.com/equineexchange/EquineExchange.ImageHosting/commit/de36584935a5bc47b3c4f732af289f75aa4581da#diff-f2c1409a20722ddaf5705add38f2fbe1

**Actual behavior:** The NuGet packages are not linkified like they are in other diffs

I've worked on the NuGet references but I don't have the time right now to look into this. I wanted to report it in case someone else might know what's causing the issue.

I've been noticing this in a lot of my recent PR diffs. Sometimes the diff will contain say 8 files that contain NuGet package references, but only 2 or 3 of them will actually be linkified. In other cases all the files are. It's happening in both my `packages.config` and `*.csproj` files.

leomoty:
As I mentioned in #449, not all files are loaded within the document body. Github generates divs with the class `.js-diff-progressive-container`, that progressive container may load more content via fetch.

The goal of this PR is to watch for changes in `.js-diff-progressive-container` and rerun octolinker init accordingly. Beware that a `.js-diff-progressive-container`, may contain other progressive containers inside, such as this commit:

https://github.com/samuelkubai/skoolspace/commit/185017f5981fe6295cd8096dbf71f61896dd8a84

Div `#files` contains two `.js-diff-progressive-container`s at DOM ready. After scrolling, another nested `.js-diff-progressive-container` is inserted when more content is downloaded.

I can write a test if we decide to follow this route.

Closes #449 

### Checklist:

### DIFF
--- a/packages/core/octo-linker.js
+++ b/packages/core/octo-linker.js
@@ -82,10 +82,27 @@ export default class OctoLinkerCore {
   }
 
   init() {
-    injection(run.bind(null, this));
+    const runCb = run.bind(null, this);
+
+    injection(runCb);
+
+    const progressiveObserver = new MutationObserver((ch, ob) => {
+      $('.js-diff-progressive-container').each((idx, element) => {
+        ob.observe(element, {
+          childList: true,
+        });
+      });
+      runCb();
+    });
+
+    $('.js-diff-progressive-container').each((idx, element) => {
+      progressiveObserver.observe(element, {
+        childList: true,
+      });
+    });
 
     $('.js-expand').on('click', event => {
-      waitFor($(event.target).closest('tbody')[0], run.bind(null, this));
+      waitFor($(event.target).closest('tbody')[0], runCb);
     });
   }
 }


stefanbuck:
I'm just catching up with your latest comment on #449 That was very well hunted down! I need some more time to think about it and will come back to you tomorrow (I'm a bit tired today).

leomoty:
No, not at all, i'll try to take a look this weekend.

stefanbuck:
I found out that `pjax` is still in use, but not everywhere. It looks like Pull Request diff views use `pjax:end` event when switching from `Conversation` tab to `Files changed` tab, but there is no `pjax:end` emitted when visiting this page directly e.g. on page reload. It also don't fire when expanding a diff or on big Pull Requests when lazy loading kicks in.

leomoty:
With that in mind, what is your suggestion?

stefanbuck:
Maybe then the `MutationObserver` unless there is a better way 😉I'll keep on exploring

stefanbuck:
I tried to follow the obfuscation javascript code form GitHub.com. From what I can tell they use `MutationObserver` too and I can't think of anything else we can do. 

I'll do a proper code review along with a manual QA session to verify that your current implementation is stable and does not break anything or has side effects. If you have time, feel free to also do a sanity test on your own.

stefanbuck:
Yesterday, I tried to come up with another MutationObserver solution and I started from scratch. However, at the end it was almost like your's 😉 

```diff
diff --git a/packages/core/octo-linker.js b/packages/core/octo-linker.js
index 6200c15..0046ce2 100644
--- a/packages/core/octo-linker.js
+++ b/packages/core/octo-linker.js
@@ -76,6 +76,16 @@ function waitFor(el, cb) {
   });
 }

+function watch(viewSpy) {
+  $('.js-diff-progressive-container, .js-diff-load-container').each(
+    (idx, element) => {
+      viewSpy.observe(element, {
+        childList: true,
+      });
+    },
+  );
+}
+
 export default class OctoLinkerCore {
   constructor(options) {
     initialize(this, options);
@@ -86,6 +96,12 @@ export default class OctoLinkerCore {

     $('.js-expand').on('click', event => {
       waitFor($(event.target).closest('tbody')[0], run.bind(null, this));
+
+    const viewSpy = new MutationObserver(() => {
+      watch(viewSpy);
+      run(this);
     });
+
+    watch(viewSpy);
   }
 }
```

I think there are still a few things that we should improve. 

- The MutationObserver callback should not invoke `run` if there are only deletions.
- We need to bind event listeners for `.js-expand` on the new dom nodes as well. Maybe this should become an delegate so we don't bind events for every single node.

leomoty:
@stefanbuck what if someone deleted a file with stuff we can octolink?

stefanbuck:
I'm talking about DOM node deletions not actual diff deletions, sorry for this confusion.

Fore example, once the content is loaded it replaces the placeholder node with the actual content which results two mutations and we can ignore the delete mutation.

leomoty:
@stefanbuck oh true, thanks for clarifying.

stefanbuck:
Hey @leomoty, good news on this front. This evening I made good progress and as far as I can it covers all scenarios. It will be part of the next major release which is scheduled for mid April. Once I'm done with testing from my side, I can share a build with you if you want.

leomoty:
Hey @stefanbuck, great to hear, sure, I'd love to test.

stefanbuck:
@leomoty please disable or deinstall the current OctoLinker extension to avoid clashes. I uploaded the beta version as a separate extension which you can install from the chrome webstore  https://chrome.google.com/webstore/detail/octolinker-beta/kcfjenhpbkggineieeknamgioeaogdmm


### Highlights
- Insert real href-links which the browser understands and applies the typical options you'd expect on. So you're able to e.g. right-click and open links in a new tab/window or middle-click the links, etc.
- Prefetch all links in the background. This is actually the key which enables all the existing improvements.
- Use the GitHub API to fetch a tree structure of the repo to resolve relative links much faster and accurate than before.
- Loads relative files without reloading the page

Let me know if you have any issues

leomoty:
Hey @stefanbuck, great work.

* The proper links are pretty good and works as expected.

* The API call gets cached! This alone is a huge win.

* I am not sure I understand the "loads relative files", can you give me an example?

* Didn't "Expand All" work before?

https://github.com/cockpit-project/cockpit/pull/9943/files

image

Both cockpit and react are not octolinked after you expand it, doesn't matter if single view/split. Didn't we "fix" this before? Been a while since I last tried finding any issues.

It has been a while since I last poked with Octolinker, mostly I use it as my daily driver which is pretty satisfying.

stefanbuck:
> I am not sure I understand the "loads relative files", can you give me an example?

In the current version of OctoLinker, following [this import](https://github.com/mvollmer/cockpit/blob/0897c1bb76dd352f8587c438ab428f6ead503c40/pkg/storaged/iscsi-panel.jsx#L23) cause a full page reload whereas the beta uses GitHub internals to load the next page with ajax which is meant to be faster. 

demo


> Didn't "Expand All" work before?

Looks like I just implemented half of it. As you've pointed out in https://github.com/OctoLinker/OctoLinker/issues/449, some blobs are not loaded with the document and I added support to link those, but not for the expand functionality 🤕 I'll look into this as well. Thanks for pointing that out. 

> It has been a while since I last poked with Octolinker, mostly I use it as my daily driver which is pretty satisfying.

Glad to hear and I hope the upcoming changes help you enjoy OctoLinker a bit more 😉

leomoty:
Also, I kinda now why it "worked", I mean, this PR was meant to fix it, which you said you were going to do an improved version with proper support. 👍 

Overall pretty pleased with the beta, I'd still love to tackle the split view somehow, if you are up for discussion after v5 lands.

stefanbuck:
True, I'll try to make it part of v5 as well.

leomoty:
If I can help in any form, do tell.

leomoty:
Hey @stefanbuck, it seems that the beta already transform it to a link, however it has no href whatsoever:

image

I am using the same diff view I added to #510 to test.

stefanbuck:
Well, isn't this a different issue? There is no expand functionality involved.

stefanbuck:
Just noticed that this file moved to another location which explains why the left hand side isn't resolvable. OctoLinker resolves files relative to the current location, not the previous one or both. That's something I've noticed a few weeks ago as well. Probably not worth the effort to handle this special case.

leomoty:
Right that was a different issue, I brought these two:
* The one this PR wanted to tackle (#449), which I seem you handled pretty closely to your implementation at this PR;
* The split view eating lines because it treated everything as a contiguous line, instead of two actual code blocks. Thus not being able to octolink it since it was broken per regex terms (#510).

I got curious and was testing the latter.

leomoty:
On top of that, that makes sense, if it is resolving compared to master, it explains why href is absent. I also agree that it is not worth the effort.

stefanbuck:
I uploaded a new version to the beta channel `4.91.0` which should be available within 30min. It handles expanding diff views.

leomoty:
Funny thing, already updated, and working as expected.

leomoty:
Legitimate question, should this be an octolinker-link?

image

stefanbuck:
Are you a professional QA engineer 😆This is definitely a bug.

leomoty:
Just pretty good at breaking things, actually. Great progress on expando!

leomoty:
Also just saw #549, if you want, we can close this and continue there. Since the v5 already does this.

stefanbuck:
Happy to close it 😄

PR #518, opened 2018-11-02
Babel 7 update

stefanbuck:
I used `npx babel-upgrade --write --install` for the basic changes, but one  additional change was needed to make it working see comment below


babel.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
```
stefanbuck: I had to rename the file to bypass a breaking change in babel 7 https://github.com/facebook/jest/issues/6053#issuecomment-383632515

stefanbuck:
Mhhh locally it works, but on travis it doesn't work. Looks like babel is not running over `_jestsetup.js` 😕 

```js
 /home/travis/build/OctoLinker/OctoLinker/_jestsetup.js:184
    import fs from 'fs';
           ^^
    SyntaxError: Unexpected identifier

```

PR #529, opened 2018-12-16
Fetch live resolver urls in background 

stefanbuck:
This is one (of a few) changes which are needed to implement #277.

On app startup it prefetch all live resolver urls at once in the background by calling a bulk endpoint. For details check out https://github.com/OctoLinker/live-resolver/pull/59. If a user clicks on a link, the click handler will check an internal cache first. If the requested resource isn't available it will fallback to the old behaviour. 

With this Pull Request I want to verify that our live resolver server is able to handle such traffic increase. The timing for this is perfect. It's one week before Christmas which I assume is already a less busy week than any other week. Also in case something goes really bad, it won't affect that many people.

### DIFF
--- a/packages/core/click-handler.js
+++ b/packages/core/click-handler.js
@@ -11,10 +11,17 @@ const LINK_SELECTOR = '.octolinker-link';
 const $body = $('body');
 let matches;
 
-function openUrl(url, newWindow = false, newWindowActive = true) {
+function openUrl(event, url) {
   if (!url) {
     return;
   }
+
+  const newWindow =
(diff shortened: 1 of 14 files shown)


stefanbuck:
Gangsta merge 😉

stefanbuck:
So, as expected, the server quickly reached a critical memory usage. This is caused by our current caching strategy. We use [mem](https://github.com/sindresorhus/mem) which worked well in the past, but maybe it isn't suitable with our new requirement. I'll investigate. 

image

PR #542, opened 2019-02-02
Add helper-github-api package

stefanbuck:
Another PR along the way to refactor OctoLinker to the next major version. Soon we will use this package to resolve relative files using the github tree api https://developer.github.com/v3/git/trees/


packages/helper-github-api/index.js
```diff
@@ -0,0 +1,3 @@
+import fetchTree from './fetch-tree';
```
stefanbuck:
```suggestion
const http = require('http')
```

Issue #530, opened 2018-12-17
Start processing the link on hover / speculatively


rauchg:
In most cases, we should not have to see the popover and read "Processing". That should be more of a worst case scenario.

stefanbuck:
I agree with you. This has been raised a few times already #277. I'm currently working on a fix. A first version to prefetch external links in the background went out a few hours ago. I'm using shadow loading to see the real production load on my Heroku plan. You still get the popover, but you will be redirected immediately after a click on e.g. `require('lodash')`. Relative files are not affected by this change yet.

rauchg:
@stefanbuck if it fits a serverless architecture, I'd love to host you on [Now](https://zeit.co/now) and sponsor this project. I love it.

stefanbuck:
Thanks @rauchg that's a generous offer. 

With a few tweaks the service should be serverless ready. It's a bunch of handlers and a few static assets to support java resolution. We use serverless for some internal tooling at work, but I never used serverless in production. Therefore I have some thoughts.

I'm little bit worried how to achieve caching in a serverless environment. Right now, the service stores all resolved url in memory. This was working fine in the past where the service was just requested on click. With the recent mentioned changes every browser navigation is prefetch links in the background. As expected this put a lot of traffic on the service and caused some serious memory issues. 

Also I wonder about the cost factor putting a service into serverless which 
(according to New Relic) handled 36.2K request within the last 24 hours. 

However, I would like to give `Now` a try. I'm really attracted by the fact to be able to deploy the service in different regions. Right now it's hosted in the US wich isn't ideal.

josephfrazier:
Oh man, I would love to be able to have this on Now! I've been keeping an eye on Now for a while, and I keep liking it more and more, but haven't actually used it for anything.

stefanbuck:
I've gotten the bare minimum (npm only) working here https://octolinker.now.sh/?package=lodash There is still some work left to do, but once it's complete I'll send a subset of users over to see how the service behaves in a serverless environment.

stefanbuck:
This will be fixed in the next major release which is scheduled for end of April. In case you don't want to wait, a beta version is available here https://github.com/OctoLinker/OctoLinker/issues/549

Thanks again for sponsoring OctoLinker. It was fun project to move the service over to NOW.

PR #543, opened 2019-02-15
Prefetch urls in background and insert "real" anchor links

### LINKED ISSUE #277: Show status bar

stefanbuck:
The overall goal of this Pull Request is to process links in the background before you actual click on a link. Previously, OctoLinker started processing links just on click for various reasons. From a UX point of view this wasn't really great. It was slow, but more important it broke the expected behaviour of a anchor tag element. It was basically just a dumb anchor tag `<a href="#" />` and we had to re-implement things like CMD+click and middle mouse button to open the url in a new tab. Also hovering over the link to see where OctoLinker takes you wasn't possible. Over the last few months, I did a lot major refactoring which was required to shipping this feature and this change will finally ship the feature. 

In order to resolve relative links within a repo, OctoLinker is now using the GitHub API. Therefore you need to create a GitHub token for private repositories. The architecture of our OctoLinker service which is responsible to resolve registry packages also changed quite a lot and we will share some updates soon.

I still need to write a few unit + E2E tests followed by a proper Q&A session. I'm considering a beta phase for this release. If you're interested in being a beta tester react with 🚀 below and I'll come back to you. 

Fixes #277 #530


stefanbuck:
@josephfrazier I think you should be able to push straight to my branch.

josephfrazier:
Ah whoops, I was looking directly at the commit, outside the context of a PR, so I thought it might have already been merged. Will do!

stefanbuck:
Before this PR is turning into a huge monster, I'm going to merge it into a v5 branch. As far as I can tell everything works as expected. Will do some more testing later this week.

PR #624, opened 2019-07-31
Support JavaScript URL imports

MarkTiedemann:
See: https://github.com/OctoLinker/OctoLinker/issues/613

Looking for feedback, @stefanbuck.

### DIFF
--- a/e2e/fixtures/javascript/index.js
+++ b/e2e/fixtures/javascript/index.js
@@ -3,3 +3,6 @@ require('fs');
 
 // @OctoLinkerResolve(https://github.com/lodash/lodash)
 require('lodash');
+
+// @OctoLinkerResolve(https://deno.land/x/std/fs/mod.ts)
+import * as fs from "https://deno.land/x/std/fs/mod.ts";
--- a/packages/plugin-javascript/__tests__/index.js
+++ b/packages/plugin-javascript/__tests__/index.js
@@ -33,9 +33,11 @@ describe('javascript-universal', () => {
   });
 
   it("does not resolve 'https://example.org/'", () => {
+    // TODO: Currently resolves incorrectly as 'https' module,
+    // but should not resolve.
     expect(plugin.resolve(path, ['https://example.org'])).toEqual({
-      target: 'https://example.org',
-      type: 'trusted-url',
+      registry: 'npm',
+      target: 'https',
     });
   });
 
--- a/packages/plugin-javascript/__tests__/index.js
+++ b/packages/plugin-javascript/__tests__/index.js
@@ -25,6 +25,29 @@ describe('javascript-universal', () => {
     });
   });
 
+  it("resolves 'https://example.org/foo.ts' to 'https://example.org/foo.ts'", () => {
+    expect(plugin.resolve(path, ['https://example.org/foo.ts'])).toEqual({
+      target: 'https://example.org/foo.ts',
+      type: 'trusted-url',
+    });
+  });
+
+  it("does not resolve 'https://example.org/'", () => {
+    // TODO: Currently resolves incorrectly as 'https' module,
+    // but should not resolve.
+    expect(plugin.resolve(path, ['https://example.org'])).toEqual({
+      registry: 'npm',
+      target: 'https',
+    });
+  });
+
+  it("resolves 'https://example.org/foo.js' to 'https://example.org/foo.js'", () => {
+    expect(plugin.resolve(path, ['https://example.org/foo.js'])).toEqual({
+      target: 'https://example.org/foo.js',
+      type: 'trusted-url',
+    });
+  });
+
   it("resolves './modules/es6.symbol' without stripping .symbol suffix", () => {
     expect(plugin.resolve('', ['./modules/es6.symbol'])).toMatchSnapshot();
   });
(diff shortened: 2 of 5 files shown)


MarkTiedemann:
Not sure what's wrong with the snapshots. Can you take a look, @stefanbuck?

stefanbuck:
First, thanks for taking the time to submit this feature. 

The snapshot is failing on Travis, because by default Jest does not write and snapshot changes when running on CI. I checked out your branch locally and running `yarn test` adds this snapshot. 


```diff
+exports[`normaliseResolverResults converts [object Object] 1`] = `
+Array [
+  Object {
+    "registry": "npm",
+    "target": "foo",
+    "type": "registry",
+  },
+]
+`;
```

I wonder why this changed. Anyway, maybe best to just revert the snapshot change which solved the issue for me. 

I'll have a closer look at the other changes later today. 

Just one thing upfront, do you mind to add an End2End test or two for this change. It's super easy to add and described here 
https://github.com/OctoLinker/OctoLinker/tree/master/e2e Thank you

packages/plugin-javascript/index.js
```diff
@@ -59,17 +59,32 @@ export function javascriptFile({ path, target }) {
...
+  try {
+    const { protocol, pathname } = new URL(target);
+    if (protocol !== 'https:' && protocol !== 'http:') return false;
+    const ext = extname(pathname);
+    if (ext !== '.js' && ext !== '.ts') return false;
+    return true;
+  } catch (err) {
+    return false;
+  }
+}
```
stefanbuck: Maybe we can simplify this by just checking if `target` is a valid url. It's not 100% inline with the spec of es modules , but that way it would work for similar cases as well. What do you think?

MarkTiedemann: I think checking for `:http(s)` as the protocol and `.{js,ts}` as the extension is pretty much the bare minimum of checks that need / should be done here. For example, `new URL("a:b")` is a valid URL (protocol `a:`, path `b`).

packages/plugin-javascript/__tests__/index.js
```diff
@@ -25,6 +25,29 @@ describe('javascript-universal', () => {
...
+  it("resolves 'https://example.org/foo.ts' to 'https://example.org/foo.ts'", () => {
+    expect(plugin.resolve(path, ['https://example.org/foo.ts'])).toEqual({
+      target: 'https://example.org/foo.ts',
+      type: 'trusted-url',
+    });
+  });
+
+  it("does not resolve 'https://example.org/'", () => {
+    // TODO: Currently resolves incorrectly as 'https' module,
+    // but should not resolve.
```
MarkTiedemann: Not sure how to fix this. ^

MarkTiedemann:
> Anyway, maybe best to just revert the snapshot change which solved the issue for me.

Sure, will do.

> do you mind to add an End2End test or two for this change

No, I didn't know you had e2e tests. Will add some tomorrow!

> but apart from that this looks good to me.

Awesome. :)

MarkTiedemann:
I deleted the snapshot file, then re-ran `yarn test` - but there's no diff. Am I missing anything on how snapshots work? I don't have any experience with them.

MarkTiedemann:
I added 1 simple e2e test - I hope that's fine.

stefanbuck:
I added two addtional commits on top. The first is reverting the snapshot (I have no idea what is happening on your end). Anyway, let's see if this work on Travis. The other is using [is-url](https://github.com/segmentio/is-url) which also enabled to fix the other test.

stefanbuck:
Mhhh still failing 😕 The latest master is passing so something must be odd with this branch. I'll give it another try tomorrow. What do you think about the other changes I made?

MarkTiedemann:
> Mhhh still failing 😕 The latest master is passing so something must be odd with this branch. I'll give it another try tomorrow.

That is weird indeed. At the weekend, I would have time to debug as well.

> The other is using `is-url` which also enabled to fix the other test.

I think using `is-url` is fine, but I'd prefer `new URL()` so we can ship less code (with [94%](https://caniuse.com/#search=url), browser support is pretty good).

stefanbuck:
Looks like the last commit 4d132be fix it. My other commit form yesterday 0cd1aae was intended to do the same, but I failed 😄 Anyway, all good now. 

Regarding the `is-url` dependency. I'm less worried about the size since it's a browser extensions and doesn't really matter if it's 10kb bigger or not. However, I went with `URL` approach and simplified it slightly. As mentioned earlier, this does not 100% match the spec since it also link just a domain, but that's fine in my opinion.

MarkTiedemann:
> Anyway, all good now.

Awesome! Thank you for fixing the snapshots!

> I'm less worried about the size since it's a browser extensions

True.

I think a better reason for using `new URL()` rather than `is-url` is that it's a standard rather than just some NPM package. So, in theory, it should be easier to understand and maintain this code.

> As mentioned earlier, this does not 100% match the spec since it also link just a domain, but that's fine in my opinion.

I agree, that's fine.

Again, thanks a lot!

MarkTiedemann:
Is this ready to be merged yet, @stefanbuck? Or is there anything I still need to do?

stefanbuck:
Nope, all good. I will release a new version later this weekend.

stefanbuck:
Thanks again for adding thia feature

e2e/fixtures/javascript/index.js
```diff
@@ -2,4 +2,7 @@
 require('fs');
 
 // @OctoLinkerResolve(https://github.com/lodash/lodash)
-require('lodash');
\ No newline at end of file
+require('lodash');
```
stefanbuck:
```suggestion
require('http')
```

PR #641, opened 2019-08-13
Limit Markdown-related style to Markdown files only

### LINKED ISSUE #639: [Question] CSS clash with refined-github
Hi,
We're working on a feature in [sindresorhus/refined-github](https://github.com/sindresorhus/refined-github). We found a `.highlight` in [helper-insert-link/style.css](https://github.com/OctoLinker/OctoLinker/blob/0174f49ed52e936dadcacc81f8cdf428d86bc950/packages/helper-insert-link/style.css#L45-L47) having `position: relative`. This selector gets applied to `.file > .data.js-blob-wrapper` on files tab of a PR, breaking the feature we're working upon.

That position relative is conflicting with a piece of our CSS as described at https://github.com/sindresorhus/refined-github/pull/2320#issuecomment-520472822.

Question: What is the purpose of this `.highlight`? Can it be removed from PR files diff and PR commit page? Or can its selector be changed so it doesn't apply to `.file`? 

(Sorry for not following issue template.)

fregante:
This should fix #639, can you ensure that it doesn't break other parts I might have not considered?

### Checklist:

### Tests

Both should remain unchanged

<blockquote>

```js
require('mem')
```

</blockquote>

https://github.com/OctoLinker/OctoLinker/pull/624/files#diff-6d7f932c7f64e54bb76f6a42949338bbR5

### DIFF
--- a/packages/helper-insert-link/style.css
+++ b/packages/helper-insert-link/style.css
@@ -41,8 +41,8 @@ body.octolinker-debug .octolinker-link {
   cursor: default;
 }
 
-/* Fix line indicator for issue code blocks */
-div.highlight {
+/* Keep line indicator inside the box in Markdown `pre` blocks */
+.markdown-body .highlight {
   position: relative;
 }
 


stefanbuck:
Great, thanks for this patch. Look good to me.

PR #662, opened 2019-09-16
Re-enable E2E tests

stefanbuck:


stefanbuck:
I'm more than happy to raise a PR for that

Issue #557, opened 2019-06-07
A script in the extension "OctoLinker" is causing Firefox to slow down


fregante:
I consistently get this message when visiting this large file:

<img width="485" alt="A script in the extension  OctoLinker  is causing Firefox to slow down" src="https://user-images.githubusercontent.com/1402241/59094859-ac6db680-8949-11e9-86f5-2cdf9766c1aa.png">

https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts

Admittedly this is more of an edge case, given the length of the file, but freezing the browser for a few seconds is never good.

Perhaps the operation can be done in the background and/or split up in multiple batches to avoid long blocking.

stefanbuck:
Wow this file length is insane! I think it's best to not execute OctoLinker at all if a file goes beyond a certain line limit. Refactroing the extension to support such edgaces seems to be overkill and I would rather spend my time with new features. 

Defining the limit is probaly the trickest task here. My gut feeling says that everyone beyone 5000 lines should be ignored. 

How does this sound to you?

bfred-it:
That sounds alright, but it might be a good opportunity to find some bottlenecks, like these long garbage collection events

<img width="1032" alt="Screenshot 2019-06-07 at 19 31 59" src="https://user-images.githubusercontent.com/1402241/59101284-117dd800-895b-11e9-95cc-c06b4f62e009.png">

It appears to be caused by this: https://github.com/OctoLinker/OctoLinker/blob/0174f49ed52e936dadcacc81f8cdf428d86bc950/packages/helper-insert-link/index.js#L2

I'm sure that there's a more efficient way to replace dom elements, e.g. looping [getTextNodes](https://github.com/sindresorhus/refined-github/blob/master/source/libs/get-text-nodes.ts)

silverwind:
I too regularily observer the browser tab hanging when opening moderately big `package.json` files. For example https://github.com/elastic/kibana/blob/master/package.json freezes my browser tab for almost 20 seconds, which I think is borderline unacceptable.

If CPU-intensive work is to be done, maybe offload it to a web worker? Thought based on the comments above, the bottleneck sounds like inefficient parsing.

bfred-it:
> freezes my browser tab for almost 20 seconds, which I think is ~~borderline~~ unacceptable.

Wow, that file is _only_ 440 lines but freezes the browser for longer than TypeScript's 18K lines file

stefanbuck:
[...]

PR #774, opened 2020-01-17
Performance improvement

stefanbuck:
OctoLinker works fine on medium size file. However, traversing the DOM for large files takes some time (especially on Firefox). In some cases this even freezes Firefox see #557 

This was caused by too many DOM lookups when applying the same RegExp over and over again to find keywords in the DOM. I've changed it so it performs the lookup against a string representation first and then updates DOM. This speeds it up significant.

### v5 (previously)

v5

### v6 (now)
v6

Based on data I was working with, this reduces processing by ~80% on large files. Small and medium files are processed ~30% faster.


packages/blob-reader/blob.js
```diff
@@ -16,32 +16,45 @@ async function fetchRaw({ user, repo, ref, path }) {
...
-  constructor({ el, path, lines, blobType, branch = undefined }) {
+  constructor({ el, path, lines, type, branch = undefined }) {
     this.el = el;
     this.path = path;
     this.lines = lines;
     this.branch = branch;
-    this.blobType = blobType; // one of: diffLeft, diffRight, snippet, full
+
+    if (this.lines.length > 5000) {
+      this.lines = this.lines.slice(0, 500);
```
jdreesen: Is this intended to be `5000` and `500`?

stefanbuck: Yes it is, but I agree the numbers are confusing. The idea is to limit OctoLinker to take only the first X lines into account for files with more than X lines. I admit those are all magic numbers I just came up with. I have seen files with ~2000 lines so I thought 5000 would be a reasonable number. Since most imports are near the top I limited it further to the first 500 lines. Maybe I should update the value to 5000 as well just for the sake of clarity. Thanks for the hint!

fregante:
I just saw those gifs; 🔥

PR #785, opened 2020-01-24
Expand .net/nuget support for .targets files and FrameworkReferences

xt0rted:
### Checklist:

- [x] If this PR is a new feature, please provide at least one example link
- [x] Make sure all of the significant new logic is covered by tests

Here's a list of files that can be used to verify these changes:

- https://github.com/SixLabors/ImageSharp.Web/blob/4a49f4d5c582d66dde963aa3eb5d88bdd5313491/src/ImageSharp.Web/ImageSharp.Web.csproj#L14-L30
- https://github.com/SixLabors/ImageSharp.Web/blob/4a49f4d5c582d66dde963aa3eb5d88bdd5313491/Directory.Build.props#L99-L105
- https://github.com/SixLabors/ImageSharp.Web/blob/4a49f4d5c582d66dde963aa3eb5d88bdd5313491/Directory.Build.targets#L19-L42
-

Closes #784

PR #787, opened 2020-01-24
Add project sdk support for .net

xt0rted:
### Checklist:

This builds on #785 so the `Sdk` attribute is also handled now.

### DIFF
--- a/e2e/fixtures/dotnet/SdkProject.csproj
+++ b/e2e/fixtures/dotnet/SdkProject.csproj
@@ -1,3 +1,4 @@
+<!-- @OctoLinkerResolve(https://www.nuget.org/packages/Microsoft.NET.Sdk) -->
 <Project Sdk="Microsoft.NET.Sdk">
 
   <ItemGroup>
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -207,3 +207,11 @@ export const NET_PROJ_PACKAGE = regex`
   (Include|Update)=${captureQuotedWord}
   .*/?>
 `;
+
+export const NET_PROJ_SDK = regex`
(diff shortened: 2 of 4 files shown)


stefanbuck:
CI stucks , most likely caused by the force push. Maybe best to recreate the commit from scratch

xt0rted:
It wasn't running before the force push. The workflow might need to be adjusted to something like this:

```yml
on:
  push:
    branches:
      - master
  pull_request:
```

That's how I set it up for [GitHub Dark](https://github.com/StylishThemes/GitHub-Dark/blob/master/.github/workflows/main.yml) so it'd handle pushes to master, PRs from local branches, and PRs from forks.

stefanbuck:
A few minutes ago I remove the `pull_request` flag https://github.com/OctoLinker/OctoLinker/commit/8feff6fbec9b5d0d6a530403a1428d76c53a4eeb since I was annoyed by the extra run 😆 Thanks for sharing this template. Will update the workflow straight away

stefanbuck:
done

xt0rted:
@stefanbuck I forgot the sdk value could contain multiple values like this `Sdk="Microsoft.NET.Sdk;Microsoft.NET.Sdk.Publish"`. Is it possible to split these and create a link for each one?

stefanbuck:
That might be tricky.  I need to look into this myself before I can say anything useful 😉

stefanbuck:
I looked into it and it's rather complicated to make this work. The current OctoLinker architecture is based on the assumption that each RegExp returns just one result. Changing this requires some larger rewrites down the line. Most likely not worth the effort.

xt0rted:
From what I know having multiple values for this isn't common. I've never seen it done in a project outside of my own so I doubt many people would hit this.

PR #789, opened 2020-01-25
Add support for .net project & file references

xt0rted:
### Checklist:

- [x] If this PR is a new feature, please provide at least one example link
- [x] Make sure all of the significant new logic is covered by tests

Adds support for project and file references inside of csproj and friends. It supports both the classic and newer sdk style formats.

Linkifying files might not be the best on large projects since you could potentially have tens of thousands of files, but that's only an issue when using the classic style. In my own testing of a project with 1,006 added links I didn't notice any additional slow down but I'm also using a core i7 6900k with 64gb of ram so this might not be the best test machine.

#### Project references

- https://github.com/aspnet/AzureIntegration/blob/f7689a10addae8a12bf84699822da155d90db788/extensions/Microsoft.AspNetCore.AzureAppServices.SiteExtension/Microsoft.AspNetCore.AzureAppServices.SiteExtension.csproj#L21-L23
- https://github.com/getsentry/sentry-dotnet/blob/ffd375caafdf7ff1a093c6e5a7e9893456e56d14/test/Sentry.Tests/Sentry.Tests.csproj#L7-L10

#### File references

- https://github.com/SixLabors/ImageSharp/blob/4973f5f160e4768af344075fe4e3fef39be26d69/src/ImageSharp/ImageSharp.csproj#L39-L205
- https://github.com/aspnet/AspNetKatana/blob/9e9c44d0b0027cb58a3252a10ea93f25ccab32af/src/Microsoft.Owin.Host.SystemWeb/Microsoft.Owin.Host.SystemWeb.csproj#L59-L167

Closes #450

PR #806, opened 2020-01-29
Add tests for plugin-dot-net-project

xt0rted:
I forgot to commit this as part of #789 😞


stefanbuck:
As you may have noticed, I tend to prefer E2E tests over unit tests. I added those kind of tests long time ago, but now with the E2E test setup I'm not sure how much value this provides. However, it does not harm just sharing my thoughts here

PR #807, opened 2020-01-29
Add support for .net global tool configs

xt0rted:
### Checklist:

**Sample files**

- https://github.com/dotnet/runtime/blob/ac68cdeeea2abbff988448fe7915f2bcf5ee3cfd/.config/dotnet-tools.json
- https://github.com/dotnet/roslyn/blob/e3a461796d3cb4917367af73778d2ab34f6812e9/dotnet-tools.json

The changes to the regex builder are needed to allow turning off the `g` flag because some tools have commands of the same name, such as in the second link, and both instances of `dotnet-format` were being linkified.

### DIFF
--- a/packages/helper-regex-builder/index.js
+++ b/packages/helper-regex-builder/index.js
@@ -1,21 +1,24 @@
 import escapeRegexString from 'escape-regex-string';
 
-function regexBuilder(key, value, groupKey) {
+function regexBuilder(key, value, groupKey, global) {
   const regexKey = escapeRegexString(key);
   const keyField = groupKey ? `("${regexKey}")` : `"${regexKey}"`;
   if (Object.prototype.toString.call(value) !== '[object String]') {
-    return new RegExp(`${keyField}`, 'g');
+    return new RegExp(`${keyField}`, global ? 'g' : undefined);
   }
 
   const regexValue = escapeRegexString(value);
   const valueField = `"(${regexValue})"`;
-  return new RegExp(`${keyField}\\s*:\\s*${valueField}`, 'g');
+  return new RegExp(
+    `${keyField}\\s*:\\s*${valueField}`,
+    global ? 'g' : undefined,
+  );
 }
 
-export function jsonRegExKeyValue(key, value) {
-  return regexBuilder(key, value, true);
+export function jsonRegExKeyValue(key, value, global = true) {
+  return regexBuilder(key, value, true, global);
 }
 
-export function jsonRegExValue(key, value) {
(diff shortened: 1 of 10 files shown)


packages/helper-regex-builder/index.js
```diff
@@ -1,21 +1,33 @@
...
+  return createRegEx(`${keyField}\\s*:\\s*${valueField}`, globalSearch);
 }
 
 export function jsonRegExKeyValue(key, value) {
-  return regexBuilder(key, value, true);
+  return regexBuilder(key, value, true, true);
+}
+
+export function jsonRegExSingleKeyValue(key, value) {
+  return regexBuilder(key, value, true, false);
```
stefanbuck: Personally I would prefer adding another argument to `jsonRegExKeyValue` instead of introducing another function. Basically how it was until https://github.com/OctoLinker/OctoLinker/commit/93d85161cd3eacac9ad467dcdd55075a2566a224#diff-b0829a6c903b7a1200ba3d8cfbf4f331

xt0rted: That's actually the exact setup I had but I switched to this thinking it'd be easier to maintain. Now it makes sense as to why  there's unit tests with a 3rd parameter for those functions even though they only have 2.

stefanbuck: Up to you, I see value in both approaches ... at the end it doesn't really matter 😉 

Regarding the tests, in #774 I had to change quite a lot so it sneaked through ... sorry for the confusion

xt0rted:
While testing this I noticed that it's turning the `//` item into a link to  nuget.org. I wasn't able to find why this was happening though since `//` clearly isn't a valid package on the site.

https://github.com/xt0rted/OctoLinker/blob/c643e7de5dcfc0b0665a54af361fe8c44f331c8c/e2e/fixtures/dotnet/dotnet-tools.json#L5

stefanbuck:
The RegExp builder returns the follow RegExp

```
("//")\s*:\s*"(@OctoLinkerResolve\(https://www\.nuget\.org/packages/dotnet\-format\))"
```

After applying the RegExp the resolver gets called  with `["//", "@OctoLinkerResolve(https://www.nuget.org/packages/dotnet-format)"]` where the first element will become the `target` variable. The url of this element is `https://www.nuget.org/packages///`.

We could add a conditional to prevent this but maybe not worth adding unnecessary complexity since it's not a real issue. What do you think?

xt0rted:
I can live with it if it's just in the test files. No need to over complicate things for an edge case.

PR #808, opened 2020-01-31
Add support for .net references

xt0rted:
### Checklist:

An example of these can be seen here https://github.com/dotnet/aspnetcore/blob/f7b5aec5d7de6f782622a2dc40de74d7dedd0a91/src/Analyzers/Analyzers/test/Microsoft.AspNetCore.Analyzers.Test.csproj#L21-L33.

One of the dependencies being linked to in that example is https://www.nuget.org/packages/Microsoft.AspNetCore.Analyzer.Testing but the package doesn't exist. Do we need to add an entry for NuGet to https://github.com/OctoLinker/api/blob/master/src/registries/config.json?

### DIFF
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -201,7 +201,7 @@ export const JAVA_IMPORT = regex`
 `;
 
 export const NET_PROJ_PACKAGE = regex`
-  <(PackageReference|DotNetCliToolReference|FrameworkReference|Reference)
+  <(DotNetCliTool|Framework|Package)?Reference
   \s+
   .*
   (Include|Update)=${captureQuotedWord}
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -201,7 +201,7 @@ export const JAVA_IMPORT = regex`
 `;
 
 export const NET_PROJ_PACKAGE = regex`
-  <(PackageReference|DotNetCliToolReference|FrameworkReference)
+  <(PackageReference|DotNetCliToolReference|FrameworkReference|Reference)
   \s+
   .*
   (Include|Update)=${captureQuotedWord}
--- a/e2e/fixtures/dotnet/sdk/Project.csproj
+++ b/e2e/fixtures/dotnet/sdk/Project.csproj
@@ -17,11 +17,17 @@
   <ItemGroup>
     <!-- @OctoLinkerResolve(https://www.nuget.org/packages/Microsoft.Extensions.Hosting.Abstractions) -->
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="3.1.1" />
+
+    <!-- @OctoLinkerResolve(https://www.nuget.org/packages/Microsoft.Extensions.Hosting.Abstractions) -->
+    <Reference Include="Microsoft.Extensions.Hosting.Abstractions" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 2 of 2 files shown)


packages/helper-grammar-regex-collection/index.js
```diff
@@ -201,7 +201,7 @@ export const JAVA_IMPORT = regex`
 `;
 
 export const NET_PROJ_PACKAGE = regex`
-  <(PackageReference|DotNetCliToolReference|FrameworkReference)
+  <(PackageReference|DotNetCliToolReference|FrameworkReference|Reference)
```
xt0rted: This could be switched to something like this but I'm not sure if it's worth it. Thoughts?

```suggestion
  <(DotNetCliTool|Framework|Package)?Reference
```

stefanbuck:
We could add NuGet support to https://github.com/OctoLinker/api/,  but in case this rarely happens I wonder if it's worth resolving all NuGet urls through the API which slows down the resolving part quite a bit. If this is the exception rather than the rule, we could also add a condition in to the related plugin.

packages/helper-grammar-regex-collection/index.js
stefanbuck: 👍 I like it

xt0rted:
I wanted to test out the actions reporter to see if it'd help with showing the test results but it doesn't seem to be working as expected so I'm removing it. I'll look into this some more though because it'd be helpful on my own projects too.

xt0rted:
> We could add NuGet support to https://github.com/OctoLinker/api/, but in case this rarely happens I wonder if it's worth resolving all NuGet urls through the API which slows down the resolving part quite a bit. If this is the exception rather than the rule, we could also add a condition in to the related plugin.

The dotnet org uses mostly pre-release packages hosted on myget.org and it looks like they have a bunch of internal packages that are hosted there but not on nuget.org. They're the only repos that I've seen setup like that but I'm sure this is also an issue for projects using internal/private feeds from myget, azure devops, and github package registry.

stefanbuck:
> I wanted to test out the actions reporter to see if it'd help with showing the test results but it doesn't seem to be working as expected so I'm removing it. I'll look into this some more though because it'd be helpful on my own projects too.

I started looking into [Problem Matchers](https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md) to report failing tests, but due to the fact this API does not support multiline output see https://github.com/hipstersmoothie/jest-github-reporter/issues/67 I paused working on it. There seems to be an interest from one of the Jest core maintainers see https://github.com/hipstersmoothie/jest-github-reporter/issues/67

stefanbuck:
In NodeJS there is a similar issue with standard library modules. Luckily there is a npm package available which lists all the core libraries so we can bail out early https://github.com/OctoLinker/OctoLinker/blob/master/packages/plugin-javascript/index.js#L79-L82. Not sure how much effort maintaining such a whitelist will be. 

Private packages are indeed an issue also for other plugins. In https://github.com/OctoLinker/OctoLinker/issues/487#issuecomment-447686999 a custom UI was suggested and I really like this idea, but I haven't thought it through yet.

xt0rted:
After looking it over that reporter was working correctly. It didn't seem it though because of how the e2e tests are being generated and run. If we generated a js file with tests instead of a json file then Jest should report errors in the right spot. That seems like overkill for now though.

stefanbuck:
I like the idea of generating the tests, because this lets you to focus or skip tests super easy which is not supported at all. On the other hand it's not really and issue so most likely not worth waiting time on it.

stefanbuck:
Mhh the CI is again failing on a forked branch. I really wonder what's the difference here. Anyway I'll go ahead and merge this and as usually master should pass 🤞

stefanbuck:
🤦‍♂ This PR depends on #811 in order to fix the failing e2e tests.

PR #809, opened 2020-02-02
Disable no-console warnings

xt0rted:
This should get rid of the warning annotations that keep showing up on each PR.


stefanbuck:
😄 Glad that you fixed it since I successfully ignored it

PR #811, opened 2020-02-02
Deduplicate matches for OctoLinker API only

stefanbuck:
As raised in https://github.com/OctoLinker/OctoLinker/pull/785#issuecomment-580069754 some references are not linked in a diff view. It turned out that the removing of duplicates was causing this issue for certain matches. Original implemented to remove duplicates before calling the OctoLinker API, this cleanup was performed on all matches regardless their type. Only matches with type  `registry` and `ping` need this cleanup.

Examples:
- https://github.com/OctoLinker/OctoLinker/pull/785/files#diff-9222904f7ff80fd69544d107fbf18b0a
- https://github.com/facebook/jest/pull/9443/files#diff-a568cf5f01a8aa0032eca9f2692ba0b0R8

### DIFF
--- a/e2e/diff-fixtures.json
+++ b/e2e/diff-fixtures.json
@@ -14,9 +14,5 @@
   {
     "url": "https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R42",
     "targetUrl": "https://github.com/eslint/eslint"
-  },
-  {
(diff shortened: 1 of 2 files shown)


stefanbuck:
I had to revert the e2e tests since expanding of diff views is not working in E2E tests. I'm pretty sure this was working at some point. Anyway, let's get master fixed and I'll address the test in a another PR

PR #836, opened 2020-03-08
Fix path check when folder name has spaces

xt0rted:
While working on some more .net support I came across a file that wouldn't resolve. The folder it's in has a space in it and the `captureQuotedWord` regex excludes spaces so I removed that but it still wouldn't resolve.

After digging into it more I found that this check was returning `false` because the file paths in `githubTree` have spaces while the value used to compare against contains `%20`.

I'm not sure if this is the correct spot to fix this, but it's a good way to see the issue.

While running this version of the code go to this url and add/remove the replace fix.

https://github.com/microsoft/uprove-csharp-sdk/blob/master/UProveUnitTest/UProveUnitTest.csproj#L78


stefanbuck:
Sounds sensible to me! Do you mind adding an E2E test to cover the new behaviour?

xt0rted:
Sure thing. When debugging this, is there a way to attach vscode or something so you can set breakpoints and step through the code? I tracked this down using a lot of `console.info()` calls and it wasn't fun.

stefanbuck:
Sorry for the late response, I was off sick. 

I can imagine how painful debugging with console statements is. In JavaScript ther is keyword called `debugger` which pauses JavaScript execution when dev tools are open see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/debugger

stefanbuck:
or are you referring to E2E tests?

xt0rted:
I was trying to debug in the browser with devtools open. I forgot about `debugger`, I'll give that a try later and try to get both of these finished up.

xt0rted:
@stefanbuck I adjusted this so it only applies to the .net file paths, otherwise it was breaking other areas like JS imports.

stefanbuck:
I agree, it's best to limit this change to .net files to avoid side effects.

xt0rted:
@stefanbuck I found that the extensions https://github.com/glebm/render-whitespace-on-github and https://github.com/sindresorhus/refined-github both break linking for files/folders with spaces in them due to showing the whitespace (which is an inserted `<span>` in each spot). These tend to run before Octolinker so our replacement can't find the string in the html. Think this is something we could try to work around?

This is what our [test file](https://github.com/xt0rted/OctoLinker/blob/d6e2ad33565fd44e40579a6f3f57707910ad087c/e2e/fixtures/dotnet/sdk/Project.csproj#L62) looks like.

image

And the html with the visible whitespace.

```html
<tr>
  <td id="L62" class="blob-num js-line-number" data-line-number="62"></td>
  <td id="LC62" class="blob-code blob-code-inner js-file-line rgh-observing-whitespace highlighted"><span data-rgh-whitespace="····">    </span>&lt;<span class="pl-ent">Content</span><span data-rgh-whitespace="·"> </span><span class="pl-e">Include</span>=<span class="pl-s"><span class="pl-pds">"</span>../foo<span data-rgh-whitespace="·"> </span>bar/test<span data-rgh-whitespace="·"> </span>file.cs<span class="pl-pds">"</span></span><span data-rgh-whitespace="·"> </span>/&gt;</td>
</tr>
```

stefanbuck:
For me this feels like an edge case. I rarely, almost never saw whitespaces in file and path names. Is this a .net thing?

xt0rted:
I noticed it in some older projects I've been working on. I don't think it's super common, but it's definitely possible and I've seen it many times over the years. I'm fine leaving this as-is.

stefanbuck:
Do you mind creating an issue for this just in case we want to fix it at some point?

PR #839, opened 2020-03-08
Persist update notification on reload

stefanbuck:
Recently, I was talking to friend and he made me aware of an issue with the update notification. The update notification appears immediately whenever the extension updates. If a user navigates to another GitHub page or performs a page reload, the notification is gone forever. With this change we persist the notification state until a user interaction happens (either close or following the link)

### DIFF
--- a/packages/helper-settings/index.js
+++ b/packages/helper-settings/index.js
@@ -24,6 +24,8 @@ export const set = async (key, value) => {
     [key]: value,
   };
 
+  Object.assign(store, data);
+
   return browser.storage.local.set(data);
 };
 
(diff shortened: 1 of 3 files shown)


packages/helper-settings/index.js
```diff
@@ -24,6 +24,8 @@ export const set = async (key, value) => {
     [key]: value,
   };
 
+  Object.assign(store, data);
```
stefanbuck: This change is needed to reflect a `set` call immediately. Before this was delayed until OctoLinker got invoked another time.

staabm:
it seems in FF the notitifaction does not disappear after clicking its "X"

stefanbuck:
Thanks for reporting, I'll look into this soon!

In the meantime, the Addons preference settings allows you to disable update notifications 

image
image

stefanbuck:
Thanks for reporting, I'll look into this soon!

In the meantime, the Addons preference settings allows you to disable update notifications 

image

Issue #857, opened 2020-03-23
Support Cargo.toml


silverwind:
**Describe the feature**
Support linking Rust crates defined in `Cargo.toml`.

**Demo**
- https://github.com/actix/examples/blob/5320117bcaeb8391dbf313bf897d978b2e588295/basics/Cargo.toml#L10 should resolve to https://github.com/actix/actix-web
- https://github.com/actix/examples/blob/5320117bcaeb8391dbf313bf897d978b2e588295/basics/Cargo.toml#L17 should resolve to https://github.com/tokio-rs/bytes

**Additional information**
Crate resolution already works in [`.rs` files](https://github.com/actix/examples/blob/master/basics/src/main.rs).

stefanbuck:
Will be part of the next release (most likely next week)

silverwind:
Nice work @xt0rted!

PR #858, opened 2020-03-24
Support for cargo.toml files

xt0rted:
Here's a rough pass and trying to add in Cargo.toml support as requested in #857. This was more of an exercise in trying to better understand how things work but I'm sure it can be cleaned up and used.

## Checklist:

### DIFF
--- a/packages/plugin-rust-cargo/index.js
+++ b/packages/plugin-rust-cargo/index.js
@@ -1,0 +1,45 @@
+import liveResolverQuery from '@octolinker/resolver-live-query';
+import * as TOML from '@iarna/toml';
+import insertLink from '@octolinker/helper-insert-link';
+import { tomlRegExKeyValue } from '@octolinker/helper-regex-builder';
+
+function linkDependency(blob, key, value) {
+  const regex = tomlRegExKeyValue(key, value);
+  return insertLink(blob, regex, this);
+}
+
+function processTOML(blob, plugin, config) {
+  let results = [];
+  const json = TOML.parse(blob.toString());
+  const toml = TOML.stringify(json.dependencies).split('\n');
+
+  toml.forEach(line => {
+    const match = /^([^=\s]+)\s*=\s*"([^"\s]+)"$/.exec(line);
+
+    if (match) {
+      const linked = linkDependency.call(plugin, blob, match[1], match[2]);
+
+      results = results.concat(linked);
+    }
+  });
+
+  return results;
+}
+
+export default {
+  name: 'Rust Cargo',
+
+  resolve(path, values, { type }) {
+    return liveResolverQuery({ type: 'crates', target: values[0] });
+  },
+
+  getPattern() {
+    return {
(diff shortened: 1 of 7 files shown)


packages/plugin-rust-cargo/index.js
```diff
@@ -0,0 +1,45 @@
...
+      results = results.concat(linked);
+    }
+  });
+
+  return results;
+}
+
+export default {
+  name: 'Rust Cargo',
+
```
stefanbuck: By adding `needsContext: true`, the [blob reader](https://github.com/OctoLinker/OctoLinker/blob/9e1caa3ed26514cb38c1c7b6a9f5fc5a7ec0195e/packages/core/app.js#L25) will fetch the full source code when the plugin is invoked on a PR split view. Otherwise the `TOML.parse` will fail if the source is divided and therfore not valid, similar to [plugin-npm-manifest](https://github.com/OctoLinker/OctoLinker/blob/49be306c492cae96930cfa16d163dc31076ecc60/packages/plugin-npm-manifest/index.js#L34) where we need a valid json string.

packages/plugin-rust-cargo/index.js
```diff
@@ -0,0 +1,45 @@
...
+function linkDependency(blob, key, value) {
+  const regex = tomlRegExKeyValue(key, value);
+  return insertLink(blob, regex, this);
+}
+
+function processTOML(blob, plugin, config) {
+  let results = [];
+  const json = TOML.parse(blob.toString());
+  const toml = TOML.stringify(json.dependencies).split('\n');
```
stefanbuck: What's the reason to convert the result of `TOML.parse` to a string again? `TOML.parse` returns a JSON object so you can iterate over it like this

```js
Object.entries(json.dependencies).forEach(([name, version]) => {
  console.log(`${name} @ ${version}`);
});
```

stefanbuck: This allows you to simplify this PR even further and remove `tomlRegexBuilder`

xt0rted: > What's the reason to convert the result of `TOML.parse` to a string again?

I was copying, pasting, and modifying until things started to work 😄

stefanbuck: That's how everything begins 😆

xt0rted:
The tests are passing but there's errors in the test log. I'm not sure if anything still needs fixing.

stefanbuck:
Are you talking about [these](https://github.com/OctoLinker/OctoLinker/pull/858/checks?check_run_id=532314135#step:8:15) erros in the E2E tests or unit tests? In the unit tests we get quite a few `TypeError` errors, but to be honest I haven't noticed them before and the current master build is producing the same output so I assume we're fine. Something we can tackle separate from this PR.

xt0rted:
I was referring to [these errors](https://github.com/OctoLinker/OctoLinker/pull/858/checks?check_run_id=532314135#step:6:27).

PR #866, opened 2020-03-31
Use some shortcuts in the versioning script

fregante:
I found some small improvements could be made, feel free to reject if you find it trivial.

Sadly [dot-json](https://github.com/maikelvl/dot-json) doesn't support merging arrays or else you could drop the `json` dependency.


package.json
```diff
@@ -11,7 +11,7 @@
     "e2e": "node ./e2e/generate-fixtures-json.js && jest --config=e2e/jest-e2e.config.js",
     "postinstall": "lerna bootstrap",
     "test:watch": "jest --watch",
-    "version": "json -I -f assets/manifest.json -e \"this.version='`json -f package.json version`'\" && git add assets/manifest.json && json -I -f packages/core/package.json -e \"this.version='`json -f package.json version`'\" && git add packages/core/package.json",
+    "version": "dot-json assets/manifest.json version $npm_package_version && dot-json packages/core/package.json version $npm_package_version && git add assets/manifest.json packages/core/package.json",
     "patch-permissions": "json -I -f dist/manifest.json -e 'this.permissions.push(\"https://api.github.com/\")'",
```
fregante: Why do you add this permission at build time?

stefanbuck: There was an issue with Firefox not being able to make request against the GitHub API without having `https://api.github.com` explicit being set

fregante:
Updating how? Dot-json can’t do that

package.json
fregante: You can include it for chrome as well

stefanbuck: but wouldn't this prompt a dialog asking for additional permissions when the update arrives?

stefanbuck:
Ohh I didn't noticed that `dot-json` is unable to add additional fields. In that case we can keep it or consider adding the permission as you suggested. I'm just concerned about asking for any additional permission which seem to be unnecessary for Chrome users.

package.json
fregante: I don’t remember, but since it’s GitHub itself it’s not a big deal

stefanbuck: Yeah, maybe I'm overthinking this. Alos it can be solved by communicating why this is needed now. Let's do it. Do you mind upgrading the manifest.json file as part of this PR?

stefanbuck:
Looks like the updated `package-lock.json` file is missing and therefore breaking the CI

stefanbuck:
Amazing, thank you @fregante

PR #868, opened 2020-04-01
Keep update notification closed when using close icon

stefanbuck:
Qhat you see is not what you get! Applies to `event.target` as well 😉 

The click handler condition to persist the state of the notification was never met, because `event.target` was pointing to the icon `<svg>` instead of the `<button>` 🤦‍♂ 

https://github.com/OctoLinker/OctoLinker/blob/bed093fbf5d20d2494cfae2514c8bf60c80e6fe7/packages/core/notification.js#L11-L14


staabm:
Thx for the fast fix

PR #869, opened 2020-04-01
Add hover card to preview to repositories

stefanbuck:
@xt0rted pointed me to this amazing idea which allows us to show a sneak preview of the linked repository using hover cards. I hope you like it as much as I do.

image


xt0rted:
That's a much nicer implementation than my hacky version.

PR #877, opened 2020-04-06
Switch to yarn workspaces

stefanbuck:
All of a sudden our CI job is failing and it seems to be related to `lerna`, although the dependency has not changed recently see https://github.com/OctoLinker/OctoLinker/runs/565631768?check_suite_focus=true#step:5:49

However, getting rid of `lerna` in favour of yarn workspaces was on my todo list since quite a while anyway


xt0rted:
Does this mean I need to install yarn now and use that instead of npm?

stefanbuck:
Yes it does

Issue #870, opened 2020-04-04
New permission https://api.github.com


stefanbuck:
Hello,

our next version of OctoLinker v6.3 will ask for one additional site permission to get read and change access to https://api.github.com!

This change **will temporarily disable OctoLinker** and you have to agree to the new permission to re-enable OctoLinker.

![](https://user-images.githubusercontent.com/1393946/78512363-25c17a00-77a4-11ea-896c-b11db6d15fe5.png)

### Why the change?

We want to be clear about permissions OctoLinker is using. Furthermore we need to ensure OctoLinker stays compliant with Google's [Web Store Policy](https://developer.chrome.com/webstore/terms). Also this change allows us to simplify our build process a little.

It's important to note, that **OctoLinker is accessing https://api.github.com already**, to fetch a file tree of the repository to make relative file lookup work, so basically nothing will change. 

If you have configured a GitHub token, OctoLinker will make requests on your behalf and we recommend you use a GitHub token to extend the rate limit for authenticated requests and enable private repository support.

This change will be part of our next release v6.3 which is planned for April 13, 2020, but we wanted to inform you well ahead of the change.

What if I have questions or concerns? We're here to help! Please leave a comment below or contact us on Twitter @octolinker.

OctoLinker Team

stefanbuck:
Another release with a permission request is needed, because I add `api.github.com` to the wrong mainfest field 🤦

Chaz6:
What is the requirement for the "change" permission on api.github.com? I couldn't find it in your description. I would not expect this extension to need to change anything.

stefanbuck:
Unfortunately there is no option to define fine grained permissions in a browser extensions. `read` access goes along with `change` permission which I agree can be confusing. OctoLinker needs to be able to access api.github.com to perform a GET request as describe above. No other requests are made. I hope this was helpful und please let me know if you have any concern

PR #894, opened 2020-04-17
Declare api.github.com as permission

stefanbuck:
While working on #870 I added https://api.github.com to the wrong array in the manifest.json file. It needs to be part of the permissions field.

This mistake is really annoying and leads to another permission ask when updating the extension next time 😞

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -28,7 +28,6 @@
     {
       "matches": [
         "https://github.com/*",
-        "https://api.github.com/*",
         "https://gist.github.com/*"
       ],
       "js": [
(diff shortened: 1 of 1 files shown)


xt0rted:
I tried this change last night and it didn't do anything for me 😕. Do changes to the manifest effect your dev instances or do you have to side-load to test them?

stefanbuck:
There is no file watcher listening for changes made in  `mainfest.json` which was most likely the reason why it didn't worked for you. I was using `yarn firefox-open` which does a proper build and it solved it for me.

PR #895, opened 2020-04-17
Replace primer.css with webext-base-css

fregante:
Fixes https://github.com/OctoLinker/OctoLinker/issues/893 using [webext-base-css](https://github.com/fregante/webext-base-css) and a little extra CSS.

- [x] Restore error via `inputField.setCustomValidity(error)`

⬆️  I don't know how to tackle that, feel free to edit this PR.

<table>
<tr>
<td valign=top>

ff-white

<td valign=top>

ff-black

<tr>
<td valign=top>

ch-white

<td valign=top>

ch-black


fregante:
The remaining lint error is [deprecated](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md):

> ##[error]  13:7   error  Form label must have ALL of the following types of associated control: nesting, id  jsx-a11y/label-has-for

packages/helper-settings/package.json
```diff
@@ -8,9 +8,7 @@
   "dependencies": {
     "linkstate": "^1.1.0",
     "preact": "^10.4.0",
-    "primer-core": "^7.0.0",
-    "primer-forms": "^3.0.0",
-    "primer-product": "^6.0.0",
+    "webext-base-css": "^1.1.0",
```
stefanbuck, line 11: [webext-base-css](https://github.com/fregante/webext-base-css) looks great 😍

stefanbuck:
Thanks for the Pull Request @fregante. According to https://stackoverflow.com/a/21458090/2121324 `.reportValidity()` does the trick to force a revalidation. It's already quite late, but I can have a look myself tomorrow.

assets/options.html
```diff
@@ -2,6 +2,7 @@
   <head>
     <meta charset="UTF-8" />
     <title></title>
+    <link rel="stylesheet" href="chrome://global/skin/in-content/common.css">
```
stefanbuck, line 5: Do you know if this is supposed to work in MS Edge as well? The readme of [webext-base-css](https://github.com/fregante/webext-base-css) mentions just Firefox. I'm asking because it looks like they added their own protocol `edge://`

fregante: Indeed, this only loads Firefox’ style since Firefox effectively doesn't support `options.chrome_style`.

I don't know how it looks in Edge, but I assume it just matches Chrome.

xt0rted:
This is what it looks like in Edge 83.0.478.10 (Official build) dev (64-bit) in Win10 

image

fregante:
> `.reportValidity()` does the trick to force a revalidation.

I think that's the wrong one, you'll need `.setCustomValidity` because `.reportValidity` only reports the error... but the browser doesn't know the error.

We used `.reportValidity()` in RG: https://github.com/sindresorhus/refined-github/commit/cb37f6d7808db5ee09889651af3f7f36ae0f3a62#diff-4bfcd665876550ba5b46897389349245L13

fregante:
> This is what it looks like in Edge 83.0.478.10 (Official build) dev (64-bit) in Win10

Not too shabby! Only issues:

- drop `width` on `html`
- the dark color doesn't match

The second point _should_ be taken care of webext-base-css, but I don't think there's currently a way to target Edge via CSS like I do for Firefox.

xt0rted:
This is it after the last updated.

image

xt0rted:
I also noticed the `yarn.lock` has some changes that weren't committed.

fregante:
> I also noticed the `yarn.lock` has some changes that weren't committed.

Indeed I need help with that too. For some reason when I tried adding the dependency inside the `helpers-settings` package yarn just locked up for minutes while `npm` installed it in seconds 🤷‍♂️

xt0rted:
I ran `yarn install` in the root of the project. I'm new to yarn but that seems like all I needed to do.

stefanbuck:
> I think that's the wrong one, you'll need .setCustomValidity because .reportValidity only reports the error... but the browser doesn't know the error.

From my understanding you need both if your are not using proper form submission like we do. I'm on it now

packages/helper-settings/SettingsForm.js
```diff
@@ -53,8 +58,11 @@ export default class Form extends Component {
     if (!response.login) {
       this.setState({
         tokenLoaded: false,
-        errorMessage: response.message || 'Something went wrong',
       });
+      this.tokenInputEl.setCustomValidity(
+        response.message || 'Something went wrong',
+      );
+      this.tokenInputEl.reportValidity();
```
stefanbuck, line 65: This solution is inspirited by https://github.com/developit/preact-mdl/blob/master/src/index.js#L873-L878 so it shouldn't be completely wrong 😉

stefanbuck:
@fregante any thought on styling the success message different when using dark mode or not?

image

packages/helper-settings/SettingsForm.js
```diff
@@ -53,8 +58,11 @@ export default class Form extends Component {
     if (!response.login) {
       this.setState({
         tokenLoaded: false,
-        errorMessage: response.message || 'Something went wrong',
       });
+      this.tokenInputEl.setCustomValidity(
+        response.message || 'Something went wrong',
```
fregante, line 63: Should the default message be more specific? Like “The token could not be validated”

fregante:
What do you think about replacing the field description with:

> ✅ Token successfully saved. 

- It doesn’t move the field down
- The green color is just given by the emoji, no need for further CSS

packages/helper-settings/SettingsForm.js
stefanbuck: Good point. Also the API response is not very descriptive `Bad credentials` in my opinion. Will change it later

stefanbuck:
The success message reserves some space in the layout to avoid form jumping when submitting. I tried out the other approach suggested by @fregante, but it felt wrong to hide the input field without showing a back button and at this point I decided for the simple approach

fregante:
> hide the input field without showing a back button

Hmm I didn’t suggest that, but replacing only the 1-line description below the field

stefanbuck:
I should read comments more carefully 😊 

Thanks for this very nice collaboration

PR #899, opened 2020-04-20
Add support for json config files

### LINKED ISSUE #552: Add support for `tsconfig`'s extend
Example: https://github.com/bfred-it/doma/blob/master/tsconfig.json

```json
{
	"extends": "@sindresorhus/tsconfig",
	"compilerOptions": {
		"outDir": ".",
		"target": "es2015",
		"declaration": true
	},
	"files": [
		"index.ts"
	]
}
```

`@sindresorhus/tsconfig` could link to https://github.com/sindresorhus/tsconfig#readme

xt0rted:
### Checklist:

This is a quick attempt at adding support for the json config format used by linting tools (they usually support js, json, and yaml but this is just for json right now). @fregante this was your request, do you know of any other json config files we could add to the list? I was looking at eslint but their format doesn't really work with this.

Closes #552 

---

## babel

- [.babelrc.json](https://github.com/AtofStryker/vue-monorepo-example/blob/57e1804b4fa140c85ac7052ca51383c3361f8d54/.babelrc.json)
- [babel.config.json](https://github.com/costaline/yr-cinema/blob/16672279153a095bd1ceb9512127e8b1ce410599/babel.config.json) - this format seems a little buggy still

## stylelint

- [single value](https://github.com/doong-jo/frontend-boilerplate/blob/8a256e90fc9c6330d8119b75ba492a2b4cd37c4e/.stylelintrc.json)
- [array of values](https://github.com/tc39/tc39.github.io/blob/49054344a33b6716ad4c29444187280c13da439a/.stylelintrc.json)

## tslint

- [single value](https://github.com/bfred-it/doma/blob/master/tsconfig.json)

### DIFF
--- a/packages/plugin-json-config/index.js
+++ b/packages/plugin-json-config/index.js
@@ -30,6 +30,7 @@ export default {
     return {
       pathRegexes: [
         /\.babelrc\.json$/,
+        /babel\.config\.json$/,
         /\.stylelintrc\.json$/,
         /tsconfig\.json$/,
       ],
--- a/packages/plugin-json-config/index.js
+++ b/packages/plugin-json-config/index.js
@@ -8,7 +8,7 @@ import liveResolverQuery from '@octolinker/resolver-live-query';
 
 function linkDependency(blob, key, value) {
   let regex;
-  
+
   if (Number.isNaN(key)) {
     regex = jsonRegExValue(key, value, true);
   } else {
@@ -28,7 +28,11 @@ export default {
 
   getPattern() {
     return {
-      pathRegexes: [/\.stylelintrc\.json$/, /tsconfig\.json$/],
+      pathRegexes: [
+        /\.babelrc\.json$/,
+        /\.stylelintrc\.json$/,
+        /tsconfig\.json$/,
+      ],
       githubClasses: [],
     };
   },
@@ -36,6 +40,8 @@ export default {
   parseBlob(blob) {
     return processJSON(blob, this, {
       '$.extends': linkDependency,
+      '$.presets': linkDependency,
+      '$.plugins': linkDependency,
     });
   },
 };
--- a/packages/plugin-json-config/index.js
+++ b/packages/plugin-json-config/index.js
@@ -1,0 +1,41 @@
+import insertLink from '@octolinker/helper-insert-link';
+import processJSON from '@octolinker/helper-process-json';
+import {
+  jsonRegExValue,
+  jsonRegExArrayValue,
+} from '@octolinker/helper-regex-builder';
+import liveResolverQuery from '@octolinker/resolver-live-query';
+
+function linkDependency(blob, key, value) {
+  let regex;
+  
+  if (Number.isNaN(key)) {
+    regex = jsonRegExValue(key, value, true);
+  } else {
+    regex = jsonRegExArrayValue(value);
+  }
+
+  return insertLink(blob, regex, this);
+}
+
+export default {
+  name: 'JsonConfig',
+  needsContext: true,
+
+  resolve(path, values) {
+    return liveResolverQuery({ type: 'npm', target: values[0] });
+  },
+
+  getPattern() {
+    return {
+      pathRegexes: [/\.stylelintrc\.json$/, /tsconfig\.json$/],
+      githubClasses: [],
+    };
+  },
+
+  parseBlob(blob) {
+    return processJSON(blob, this, {
+      '$.extends': linkDependency,
+    });
+  },
+};
--- a/packages/core/load-plugins.js
+++ b/packages/core/load-plugins.js
@@ -13,6 +13,7 @@ export { default as Homebrew } from '@octolinker/plugin-homebrew-manifest';
 export { default as HTML } from '@octolinker/plugin-html';
 export { default as Java } from '@octolinker/plugin-java';
 export { default as JavaScript } from '@octolinker/plugin-javascript';
+export { default as Jsonconfig } from '@octolinker/plugin-json-config';
 export { default as Less } from '@octolinker/plugin-less';
 export { default as NodejsRelativePath } from '@octolinker/plugin-nodejs-relative-path';
 export { default as NpmManifest } from '@octolinker/plugin-npm-manifest';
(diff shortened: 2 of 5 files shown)


packages/plugin-json-config/index.js
```diff
@@ -0,0 +1,50 @@
...
+import {
+  jsonRegExValue,
+  jsonRegExArrayValue,
+} from '@octolinker/helper-regex-builder';
+import liveResolverQuery from '@octolinker/resolver-live-query';
+
+function linkDependency(blob, key, value) {
+  let regex;
+
+  if (Number.isNaN(key)) {
```
xt0rted, line 12: When the value is an array the `key` values are numeric so `jsonRegExValue` doesn't work.

fregante:
In addition to the ones you already added (👌) there’s also `eslintrc` and any tools that also support adding their config to `package.json`, on keys like you can see [here](https://github.com/sindresorhus/refined-github/blob/master/package.json): `*.extends`, `*.plugins`, `*.rules`, `ava.require`

stefanbuck:
As @fregante already mentioned, ESLint might be interesting to cover as well, although from my experience many projects are using the `.eslintrc.js` file over `.eslintrc.json`. Would be good to know the market share of json vs js config files. 

Btw. I maintain a module called [eslint-rule-docs](https://github.com/stefanbuck/eslint-rule-docs) which was intended to become part of OctoLinker as well at some point, but I fall into the trap of trying to cover all possible eslint config files.

xt0rted:
Just pushed some more changes that adds package.json support for ava, babel, eslint, stylelint, and xo.

The issue with eslint is it seems like most of their references don't use the full package name so we're not able to link them. This probably needs its own plugin so we can handle looking those things up differently.

fregante:
This is looking great!

packages/plugin-json-config/index.js
```diff
@@ -0,0 +1,59 @@
...
+      '$.ava.require': linkDependency,
+      '$.babel.presets': linkDependency,
+      '$.babel.plugins': linkDependency,
+      '$.eslintConfig.extends': linkDependency,
+      '$.eslintConfig.parser': linkDependency,
+      '$.stylelint.extends': linkDependency,
+      '$.stylelintConfig.extends': linkDependency,
+      '$.xo.extends': linkDependency,
+      '$.xo.parser': linkDependency,
+    });
```
stefanbuck, line 57: I wonder if it makes a difference performance wise to have dedicated plugins for each config, but most likely it's just micro optimisation. I'll look into it to get a better understanding if this is worth changing or not.

stefanbuck: I checked a few different config files and this code run on average in less than 5ms 😄

stefanbuck:
Looks amazing @xt0rted. I don't think we need to add E2E tests for all of them but maybe for the most common once like for babel, eslint and config within a package.json file

xt0rted:
@stefanbuck I was adding tests in for this and trying to get the rest of the areas implemented but noticed an issue in how the links are being handled. If the matched text exists more than once in the file, such as `@babel/plugin-proposal-class-properties`, it's resulting in nested links like this:

image

This is the file I was testing against https://github.com/jassi1124/reaction-testing/blob/843f14aec572c3de4551b8e180698495ccd19458/package.json

I'm not really sure what I can do to prevent this.

stefanbuck:
Looks like a bug, which even exists for single dependencies. In the old implementation of `insert-link.js` there was a check to prevent such a behaviour, but now that we process line by line I thought this isn't needed anymore. It also exists in the current master so worth fixing it separately. I'll look into it
image

stefanbuck:
🤦 I had two versions of OctoLinker running (local and the live version from the store) and therefor OctoLinker run twice. However, I'll look how we can fix it here

stefanbuck:
Was the screenshot taken from your regular chrome instance or during E2E tests? I'm asking because it seems to work for me

xt0rted:
It was with my local dev instance from `yarn chrome-launch` and it was the only extension running.

stefanbuck:
Mhhh is it still behaving like that? I run `yarn && yarn chrome-launch` on this branch and it doesn't wrap the links many times.

xt0rted:
I have changes that weren't pushed up yet because I'm still playing with it. Give [this branch](https://github.com/xt0rted/OctoLinker/tree/json-configs-test) a try. It's still a work in progress but that's the code that was triggering the issue for me.

stefanbuck:
Strange, it's still behaving as expected. 

image

stefanbuck:
Sorry it was my fault. I assumed `yarn chrome-launch` does a build as well, which isn't the case. I use `yarn start` and load the extension as unpacked extensions via chrome://extensions/

stefanbuck:
but now it's too late for me 😉 I'll continue the hunt tomorrow

xt0rted:
I've been using a split terminal in vscode, one has `yarn start` and the other `yarn chrome-launch`.

stefanbuck:
I pushed a fix onto this branch c6dd2251c19f52b2cf7158bbd6379c8c7396efef which is fixing the issue by checking whether or not the node or parent is already wrapped. I need to think about how to test this best.

stefanbuck:
I pushed a test case to ensure links are not wrapped twice. This covers c6dd225

stefanbuck (approved):
Amazing feature @xt0rted

fregante:
Is this done? Can it be merged?

stefanbuck:
@xt0rted is unable to work on this PR in the near future. I'll have a closer look once Safari support #29 is added.

fregante:
> I'll have a closer look once Safari support #29 is added.

How is that related?

xt0rted:
I'll finish this up next week

stefanbuck:
> How is that related?

By the availability of my time 😉 I have to juggle kids, my day job and way too many side projects which is quite difficult during these crazy times.

fregante:
I hope to see this merged eventually ☺️

Yash-Singh1:
There is also [`markdownlint`](https://github.com/DavidAnson/markdownlint#optionsconfig) and [`cypress`](https://github.com/bahmutov/cypress-extends).

PR #943, opened 2020-06-04
More GitHub actions support

xt0rted:
### Checklist:

This builds on the existing support and adds links for local actions in the current repo, as well as actions that have sub actions which are located in a sub folder for the given action.

I had to hard code `master` for one of the urls because we don't have branch/commit information for it. What we could do is adjust the regex to capture the `@v1` value and use that. That should be a tag or commit which both work in the url and would be more accurate.

**Local**

https://github.com/github/codeql-action/blob/a0d60d5d9e3126a00bed89fd89cc89b7e52a56f2/.github/workflows/codeql.yml#L24-L28

**Sub folders**

https://github.com/twbs/bootstrap/blob/5faf41eb4837491fa8193910c5816efadb4dbc5a/.github/workflows/codeql.yml#L22-L25

### DIFF
--- a/e2e/fixtures/.github/workflows/ci.yml
+++ b/e2e/fixtures/.github/workflows/ci.yml
@@ -13,6 +13,6 @@ jobs:
     - uses: actions/upload-artifact@master
     # @OctoLinkerResolve(https://github.com/github/codeql-action/tree/master/init)
     - uses: github/codeql-action/init@v1
-    # @OctoLinkerResolve(https://github.com/OctoLinker/OctoLinker)
+    # @OctoLinkerResolve(https://github.com/OctoLinker/OctoLinker/tree/<sha>/)
     - uses: ./
     
--- a/e2e/fixtures/.github/workflows/ci.yml
+++ b/e2e/fixtures/.github/workflows/ci.yml
@@ -11,6 +11,5 @@ jobs:
     - uses: actions/setup-node
     # @OctoLinkerResolve(https://github.com/actions/upload-artifact)
     - uses: actions/upload-artifact@master
-    # @OctoLinkerResolve(https://github.com/github/codeql-action/tree/master/init)
+    # @OctoLinkerResolve(https://github.com/github/codeql-action/tree/main/init)
     - uses: github/codeql-action/init@v1
-    
--- a/e2e/fixtures/.github/workflows/ci.yml
+++ b/e2e/fixtures/.github/workflows/ci.yml
@@ -11,6 +11,8 @@ jobs:
     - uses: actions/setup-node
     # @OctoLinkerResolve(https://github.com/actions/upload-artifact)
     - uses: actions/upload-artifact@master
+    # @OctoLinkerResolve(https://github.com/github/codeql-action/tree/master/init)
+    - uses: github/codeql-action/init@v1
     # @OctoLinkerResolve(https://github.com/OctoLinker/OctoLinker)
     - uses: ./
     
--- a/e2e/fixtures/.github/workflows/ci.yml
+++ b/e2e/fixtures/.github/workflows/ci.yml
@@ -13,6 +13,4 @@ jobs:
     - uses: actions/upload-artifact@master
     # @OctoLinkerResolve(https://github.com/github/codeql-action/tree/master/init)
     - uses: github/codeql-action/init@v1
-    # @OctoLinkerResolve(https://github.com/OctoLinker/OctoLinker/tree/<sha>/)
-    - uses: ./
     
--- a/e2e/fixtures/.github/workflows/ci.yml
+++ b/e2e/fixtures/.github/workflows/ci.yml
@@ -11,3 +11,6 @@ jobs:
     - uses: actions/setup-node
     # @OctoLinkerResolve(https://github.com/actions/upload-artifact)
     - uses: actions/upload-artifact@master
+    # @OctoLinkerResolve(https://github.com/OctoLinker/OctoLinker)
+    - uses: ./
+    
--- a/e2e/generate-fixtures-json.js
+++ b/e2e/generate-fixtures-json.js
@@ -28,6 +28,7 @@ if (process.env.GITHUB_EVENT_PATH) {
 
 const user = username || 'OctoLinker';
 const branch = process.env.GITHUB_HEAD_REF || 'master';
+const sha = process.env.GITHUB_SHA || 'HEAD';
 const fixturesRoot = `https://github.com/${user}/OctoLinker/blob/${branch}/e2e`;
 
 console.log('Fixtures root:', fixturesRoot); // eslint-disable-line no-console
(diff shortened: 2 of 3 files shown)


packages/plugin-github-actions/index.js
```diff
@@ -8,6 +9,18 @@ export default {
       return '';
     }
 
+    if (target.startsWith('.')) {
+      const [, org, repo, , sha] = path.split('/');
+
+      return `{BASE_URL}/${join(org, repo, 'tree', sha, target)}`;
```
xt0rted, line 15: I have another PR I'm working on for Dart or C/C++ that does this same thing. I'll make sure to move this out to a helper when I get back to it.

xt0rted: I used `join()` for these because that's how it was done here and it seemed a little cleaner. Do you have a preference to how these are built?

https://github.com/OctoLinker/OctoLinker/blob/a36728e8b29f3d54a1770585041c331d262cc120/packages/resolver-relative-file/index.js#L4

e2e/fixtures/.github/workflows/ci.yml
```diff
@@ -11,3 +11,8 @@ jobs:
     - uses: actions/setup-node
     # @OctoLinkerResolve(https://github.com/actions/upload-artifact)
     - uses: actions/upload-artifact@master
+    # @OctoLinkerResolve(https://github.com/github/codeql-action/tree/master/init)
+    - uses: github/codeql-action/init@v1
+    # @OctoLinkerResolve(https://github.com/OctoLinker/OctoLinker)
+    - uses: ./
```
xt0rted, line 17: For some reason this test failed but it works correctly in the browser.

```
✕ resolves - uses: ./ to https://github.com/OctoLinker/OctoLinker (5001 ms)
```

packages/plugin-github-actions/index.js
stefanbuck: I don't mind, but I agree, your solution is more explicit.

e2e/fixtures/.github/workflows/ci.yml
stefanbuck: If I navigate to https://github.com/OctoLinker/OctoLinker/blob/936dbbc793bb8e2246a1ea7b6b34c76244dd28bd/e2e/fixtures/.github/workflows/ci.yml

this line links to https://github.com/OctoLinker/OctoLinker/tree/936dbbc793bb8e2246a1ea7b6b34c76244dd28bd/

which is inline with your implementation. Maybe we need another wildcard like `<root>` https://github.com/OctoLinker/OctoLinker/blob/master/e2e/generate-fixtures-json.js#L64

xt0rted: How do you run the e2e tests locally? I adjusted `generate-fixtures-json.js` to point to my fork & branch, but I keep getting the following output when trying to run them:

```
❯ node "c:/dev/xt0rted/OctoLinker/node_modules/jest/bin/jest.js" "c:/dev/xt0rted/OctoLinker/e2e/automated.test.js" -t "End to End tests single blob"
No tests found, exiting with code 1
Run with `--passWithNoTests` to exit with code 0
In C:\dev\xt0rted\OctoLinker
  198 files checked.
  testMatch: **/__tests__/**/*.[jt]s?(x), **/?(*.)+(spec|test).[tj]s?(x) - 37 matches
  testPathIgnorePatterns: C:\\dev\\xt0rted\\OctoLinker\\e2e\\ - 186 matches
  testRegex:  - 0 matches
Pattern: c:\\dev\\xt0rted\\OctoLinker\\e2e\\automated.test.js - 0 matches
```

The `fixtures.json` was created, but the tests in `automated.test.js` don't run.

stefanbuck: There is a special jest config for e2e tests `--config=e2e/jest-e2e.config.js` which is used when running `yarn e2e`

e2e/generate-fixtures-json.js
```diff
@@ -18,11 +18,14 @@ const async = require('async');
 // ]
 
 let username;
+let sha = 'HEAD';
```
stefanbuck, line 21: This won't work when running `yarn e2e` local. I resist to call out the git binary to get the latest sha, because this doesn't work under all conditions for example the latest commit isn't pushed yet.

stefanbuck: Maybe we should consider not adding tests for relative links in workflow files. What do you think @xt0rted ?

xt0rted: I'm ok with that. Should I remove that test and rebase this or do you want to merge it with the fixup commits?

stefanbuck: I think using an env var doesn't solve the actual issue, because the sha will be different all the time, right? Also this would fail when running on a fork right? Let's just remove that test and get this merged.

xt0rted: When these run they pull whatever is in the current branch, so the sha will change on each run. If the test only ran when `GITHUB_ACTION` was `true` then we'd also be able to get the sha from `GITHUB_SHA`, but that won't work with the current test setup. Maybe in the future, but not now.

packages/plugin-github-actions/index.js
```diff
@@ -8,6 +9,21 @@ export default {
...
+    }
+
+    const [org, repo, ...folder] = target.split('/');
+
+    if (folder.length) {
+      return [
+        `{BASE_URL}/${join(org, repo, 'tree', 'main', ...folder)}`,
```
xt0rted, line 22: This is because GitHub is starting to rename their default branches from `master` to `main` and I think this is supposed to become the site's default for new repos soon.

e2e/generate-fixtures-json.js
stefanbuck: A while ago I came across the fact that the env var `GITHUB_SHA` is different depending on whether your workflow runs on push or pull_request events.

Pull request event:  Last merge commit on the GITHUB_REF branch
Push event: Commit pushed, unless deleting a branch (when it's the default branch)

https://help.github.com/en/actions/reference/events-that-trigger-workflows

xt0rted:
This should be good now.

packages/plugin-github-actions/index.js
stefanbuck: Do you mind if we address this separately? There are a few places where master is hardcoded and I think this needs some additional thoughts

xt0rted: Without this the test won't pass because the [github/codeql-action](https://github.com/github/codeql-action) repo uses `main` now.

stefanbuck: 😄  Ahh I see, thanks for the hint.

xt0rted: Opened #974 for this.

PR #975, opened 2020-07-13
Use toast to show app notifications

stefanbuck:
First attempt to tweak the app notification to become less intrusive. This toast message appears in the bottom left corner and therefore doesn't cause the page jump when showing the notification. 

## Screenshots

Screenshot 2020-07-13 at 23 03 58

Screenshot 2020-07-13 at 23 15 27

Screenshot 2020-07-13 at 23 14 48


RyanCopley:
I like this approach much better. Thank you for taking time out of your schedule to put this together.

xt0rted:
Looks like I was wrong, the toast design does come from github. Now if I could only get the site to show me one.

stefanbuck:
It's part of the notification page when clicking the "Mark as done" button

Large GIF (888x470)

stefanbuck:
> I'm not sure if this is by design but clicking anywhere on the toast closes it, not just the X button.

This maybe caused by the click listener which is supposed to close the toast when following a link from there. I'll look into it.

stefanbuck:
@xt0rted I pushed a single fixup commit with a bunch improvements.

- No global click listener anymore 
  - Does not close the toast message when clicking anywhere on the toast
  - Custom behaviour like opening the settings page is handled with a new `clickHandler` prop
- Using the x icon to close API related messages, does set a timestamp to prevent messages from appearing all the time.
- Inline release notes to avoid unnecessary use of API calls. 
- Make use of the `webextension-polyfill` where `chrome.` was still used
- Added OctoLinker mascot to the notification 


Screenshot 2020-08-02 at 00 10 41

RyanCopley:
That logo addition is adorable.

packages/user-interface/index.js
```diff
@@ -1,34 +1,48 @@
...
-    error: 'flash-error',
+    info: 'Toast--info',
+    success: 'Toast--success',
+    warn: 'Toast--warning',
+    error: 'Toast--error',
+  };
+
+  const icons = {
+    info:
+      '<svg style="pointer-events: none;" class="octicon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" width="16" height="16"><path fill-rule="evenodd" d="M8 1.5a6.5 6.5 0 100 13 6.5 6.5 0 000-13zM0 8a8 8 0 1116 0A8 8 0 010 8zm6.5-.25A.75.75 0 017.25 7h1a.75.75 0 01.75.75v2.75h.25a.75.75 0 010 1.5h-2a.75.75 0 010-1.5h.25v-2h-.25a.75.75 0 01-.75-.75zM8 6a1 1 0 100-2 1 1 0 000 2z"></path></svg>',
```
xt0rted, line 18: The svgs GitHub uses on the notification page don't have `point-events` set, I don't think that's needed if the click handler is moved to the close button.

GitHub's also setting `aria-hidden="true"` on these.

packages/user-interface/index.js
```diff
@@ -1,34 +1,48 @@
...
-  document.getElementById('js-flash-container').append(el);
-
-  return el;
-};
-
-export const removeAllNotifications = () => {
-  document
-    .querySelectorAll('.js-octolinker-flash')
-    .forEach((el) => el.remove());
+  document.body.append(el);
```
xt0rted, line 41: We could insert these right into the alert container GitHub uses:

```js
document.querySelector('#js-flash-container').append(...el.children);
```

This element should exist on every page, I haven't seen one without it.

They do some special handling of alerts but I'm not really sure what it's for. It might be to swap them out if the pjax response is an alert, in which case it might not be a good idea to put them in this container. You can see this in the `github-bootstrap.js` file by searching for `js-flash-container'`.

stefanbuck: The reason why I moved away from this `#js-flash-container` container was to ensure the toast keeps showing even GitHub removes this container one day. To be fair, OctoLinker is using this container since a couple of years, but you never know. Also, as far as I tested this does not cause any issues when switching between pjax pages. I thinks it's best to rely as little as possible on the GitHub DOM. That said, in this PR I use quite a few CSS classes from GitHub which can change as well 🤷‍♂️ 

Unless you have strong feelings about this container, I tend to keep it as it is.

packages/user-interface/index.js
stefanbuck: The way how we set the click handler requires us to set this css property, otherwise `event.target` is the `svg`-tag

packages/user-interface/index.js
```diff
@@ -1,34 +1,48 @@
...
+  el.addEventListener('click', function (event) {
+    clickHandler(event);
+    el.remove();
+  });
```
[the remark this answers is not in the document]
stefanbuck: > The notifications kept dismissing 

Sorry my fault, I changed the implementation without verifying the behaviour again.

Your code suggestion works for the close button, but keep in mind that most toast messages contain other links, like a link to the settings page. Without introducing a click listener for every single link, this global listener seems to be the best way to go. Does this make sense?

packages/user-interface/index.js
stefanbuck: I just checked and GitHub is appending toast messages to the body as well.
image

packages/user-interface/index.js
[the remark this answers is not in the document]
xt0rted: I didn't realize clicking on those links should close the toast.

packages/user-interface/index.js
xt0rted: I thought they were adding them to the alert container but I guess not. As long as they work leaving them out is the better way since their JS looks to clear the alerts if a new one is sent via pjax.

packages/core/notification.js
```diff
@@ -2,22 +2,11 @@ import * as storage from '@octolinker/helper-settings';
...
-    event.target.classList.contains('js-hide-new-version') ||
-    event.target.classList.contains('js-flash-close-update-info')
-  ) {
-    storage.set('showUpdateInfo', false);
-    if (notificationEl) {
-      notificationEl.remove();
-    }
-  }
-});
+const releaseDescription = 'Test';
```
xt0rted, line 5: Every time the new version notification shows it's prefixed with `Test - ` as you can see in my previous screen shots. Should this be changed/removed?

xt0rted:
This looks good now, only the close button or a link in the toast closes it. I think the new version message needs to be adjusted before this can be merged though.

packages/core/notification.js
stefanbuck: The idea was rather than to make an API call to bake the release description into the release version itself. Obviously this requires updating this string, which can slip through easily. Let me see if I can find a way to fail the build in such a case.

stefanbuck:
> I think the new version message needs to be adjusted before this can be merged though.

True, will update the message before releasing a new version

packages/core/notification.js
stefanbuck, line 5:
```suggestion
const releaseDescription = '';
```

PR #1033, opened 2020-09-21
Add PHP manual support

stefanbuck:
[](https://github.com/OctoLinker/OctoLinker/pull/1033/checks) [Pull Request Badge](https://pullrequestbadge.com/?rel=badge)

### Checklist:

Support for PHP is limited in OctoLinker, and besides making dependencies clickable in a `composer.json` we don't do any other linking. However, providing better PHP support is on my list for quite some time now. This PR links PHP core classes to the related php.net manual page

```php
<?php

use Closure; // => https://www.php.net/manual/en/class.closure.php
use Exception; // => https://www.php.net/manual/en/class.exception.php
use ArrayAccess; // => https://www.php.net/manual/en/class.arrayaccess.php
use function is_array; // => https://www.php.net/manual/en/function.is-array.php
use function preg_last_error; // => https://www.php.net/manual/en/function.preg-last-error.php

?>
```

Demo:

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Container/Container.php

### DIFF
--- a/e2e/fixtures/php/foo.php
+++ b/e2e/fixtures/php/foo.php
@@ -9,6 +9,12 @@ use Exception;
 // @OctoLinkerResolve(https://www.php.net/manual/en/class.arrayaccess.php)
 use ArrayAccess;
 
+// @OctoLinkerResolve(https://www.php.net/manual/en/function.is-array.php)
+use function is_array;
+
+// @OctoLinkerResolve(https://www.php.net/manual/en/function.preg-last-error.php)
+use function preg_last_error;
+
 use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\Contracts\Container\Container as ContainerContract;
 
--- a/e2e/fixtures/php/foo.php
+++ b/e2e/fixtures/php/foo.php
@@ -1,0 +1,15 @@
+<?php
+
+// @OctoLinkerResolve(https://www.php.net/manual/en/class.closure.php)
+use Closure;
+
+// @OctoLinkerResolve(https://www.php.net/manual/en/class.exception.php)
+use Exception;
+
+// @OctoLinkerResolve(https://www.php.net/manual/en/class.arrayaccess.php)
+use ArrayAccess;
+
+use Illuminate\Contracts\Container\BindingResolutionException;
+use Illuminate\Contracts\Container\Container as ContainerContract;
+
+?>
--- a/packages/blob-reader/fixtures/github.com/issue/code.html
+++ b/packages/blob-reader/fixtures/github.com/issue/code.html
@@ -3,7 +3,6 @@
 
 
 
-
 <!DOCTYPE html>
 <html lang="en">
   <head>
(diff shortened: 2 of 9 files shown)


packages/plugin-php/index.js
```diff
@@ -0,0 +1,26 @@
+import { PHP } from '@octolinker/helper-grammar-regex-collection';
+import liveResolverQuery from '@octolinker/resolver-live-query';
+
+export default {
+  name: 'PHP',
+
+  resolve(path, [target]) {
+    if (target.includes('\\')) {
+      return;
```
stefanbuck, line 9: I'm not a PHP developer, so I don't know if there is a case where core class contains a slash as well ... I'll investigate

staabm:
maybe it makes sense to use https://github.com/JetBrains/phpstorm-stubs as a datasource.

this repo contains information for native php functions/classes and also descriptions etc.
its used by the native phpstorm IDE autocompletion feature, so should be pretty up 2 date.

wouldn't it be usefull if ocotlinker would provide an in-github.com overlay which renders the description of the native method for quick lookups?

stefanbuck:
Thanks for the suggestion @staabm. This repo looks interesting, but I struggle to figure out how to use it for this use case. For example, `ArrayAccess` seems to be missing here. On the other hand, the current approach is quite nice, because we don't have to maintain a list. For Java package we crawl https://spring.io to generate such a mapping list, but this is quite tricky to maintain. The ping request is made from the [OctoLinker API](https://github.com/OctoLinker/api) and cached on the server so performance wise it shouldn't make a big difference. 

The overlay is an interesting idea, but we should stick to the current OctoLinker UX and don't overcomplicate things. If we decide to add such an overlay it should work for other languages too. Feel free to open a separate issue, but I can't promise anything. 

There is no plan to support other linking beside the `use` and other import statements, mainly because GitHub is working on this with code navigation already.

staabm:
I see, thx for your reply.

let me add a few more `use`-use-cases which php supports.

using of constants (which might be part of the php std lib): https://github.com/thecodingmachine/safe/blob/f34781c8a9b468954397789e244f02c55ffb745f/lib/special_cases.php#L10

using of functions (which might be part of the php std lib): https://github.com/vimeo/psalm/blob/3015aca2df8b912b0dacc2d7e6ffa79de222440a/src/Psalm/Internal/Analyzer/Statements/Block/TryAnalyzer.php#L16-L23

stefanbuck:
I added support for functions, but constants are tricky. Other than functions and classes, there is not dedicated page for every constant. For example: `PREG_NO_ERROR` is a return value of https://www.php.net/manual/en/function.preg-last-error.php. I wonder if this is just nice to have or actual useful?

staabm:
can I somehow beta test the feature?

stefanbuck:
If you want to test the feature, you have to checkout this PR and follow [contribution guide](https://github.com/OctoLinker/OctoLinker/blob/main/CONTRIBUTING.md). `yarn chrome-open` open is meant for development, so if you want to test the feature over a day or so I recommend doing a build as described [here](https://github.com/OctoLinker/OctoLinker/blob/main/CONTRIBUTING.md#user-content-quick-start:~:text=Chrome%20(Long%20Version)%3A).

staabm:
Hmm ok. Will wait for the release then. 

No interessted in setting up all the tooling etc

packages/blob-reader/fixtures/github.com/issue/code.html
```diff
@@ -18,48 +17,65 @@
 
 
 
-  <link crossorigin="anonymous" media="all" integrity="sha512-InGiw0uqrfrV1EXLEoQR8zhoNQs4RhOXAZIoTEr+QaNPzbnLcqgPASwPZOSxhLC/ydy1tqek+nufQayviU4NRA==" rel="stylesheet" href="https://github.githubassets.com/assets/frameworks-c0ad6862340d6d1a2594b40111c17bf7.css" />
-  <link crossorigin="anonymous" media="all" integrity="sha512-q0uE9UAtjRgctFaLFVvUGVtXC6ggCD9Qcg86uwRaLx2RdC7FIE+hPfB4YDaM8lCecPVMwsxDyvqkXI2XEKDCOA==" rel="stylesheet" href="https://github.githubassets.com/assets/site-d2ea9f95eb7121ca3a5f9571e0012f11.css" />
-    <link crossorigin="anonymous" media="all" integrity="sha512-JedmoBWaUotKU3Fc6fDpBbcfTTMhoAiSb4IPTqwmqhD3pBHWUtHDyYYgVGuNwIKt08sVud5nLKZCb8DBdPvtNQ==" rel="stylesheet" href="https://github.githubassets.com/assets/github-0c16774092b1ce5288b91585e3d6617b.css" />
+  <link crossorigin="anonymous" media="all" integrity="sha512-QLgu+HpQvHV00FunhsB3ueQCfgM7dkyMXKp1BrwCifVinFow2alu90jbiIbSdyn4/rXsaRo3otCqeYgmRmTCEA==" rel="stylesheet" href="https://github.githubassets.com/assets/frameworks-40b82ef87a50bc7574d05ba786c077b9.css" />
```
stefanbuck, line 20: All changes made in this file are due to the fact that I added a new snipped to https://github.com/OctoLinker/testrepo/issues/2. I ran `./packages/blob-reader/scripts/update-fixtures.sh` and committed only this file to keep the overall diff of this PR small. Interesting to see all the markup changes

packages/plugin-php/index.js
staabm: There are no namespaced classes in the php std lib

packages/plugin-php/__tests__/index.js
```diff
@@ -0,0 +1,28 @@
+import { PHP_FUNC } from '@octolinker/helper-grammar-regex-collection';
+import php from '../index';
+
+describe('rust-crate', () => {
```
xt0rted, line 4: Rename to something php related?

```suggestion
describe('php', () => {
```

xt0rted (changes requested):
Looks good to me. Only change I can see is the unit test name.

e2e/fixtures/php/foo.php
```diff
@@ -0,0 +1,21 @@
...
+// @OctoLinkerResolve(https://www.php.net/manual/en/class.arrayaccess.php)
+use ArrayAccess;
+
+// @OctoLinkerResolve(https://www.php.net/manual/en/function.is-array.php)
+use function is_array;
+
+// @OctoLinkerResolve(https://www.php.net/manual/en/function.preg-last-error.php)
+use function preg_last_error;
+
+use Illuminate\Contracts\Container\BindingResolutionException;
```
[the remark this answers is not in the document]
stefanbuck: Great idea. I'll look into it later this week

xt0rted (approved):

packages/helper-grammar-regex-collection/index.js
```diff
@@ -110,6 +110,21 @@ export const RUST_CRATE = regex`
   (?<$1>[^:;\s]+)
 `;
 
+export const PHP = regex`
+  ${diffSigns}
```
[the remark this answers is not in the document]
stefanbuck: In the `Files changed` tab on a pull request, a plus or minus signs is rendered to indicate whether a line is an addition or deletion. Note, the refined-github extension removes it. It was introduced long time ago and I think you're right we should expand this to others as well. 

https://github.com/OctoLinker/OctoLinker/pull/409 https://github.com/OctoLinker/OctoLinker/pull/410

packages/blob-reader/fixtures/github.com/issue/code.html
stefanbuck: Yes, I'm so glad that OctoLinker is not tight to the DOM anymore. At the beginning OctoLinker stop working every now and then. The regular expression approach and https://github.com/padolsey/findAndReplaceDOMText is the key why OctoLinker is stable since many years

PR #1034, opened 2020-09-23
Support extensionless less imports

xt0rted:
[](https://github.com/OctoLinker/OctoLinker/pull/1034/checks) [Pull Request Badge](https://pullrequestbadge.com/?rel=badge)

### Checklist:

Found this while setting up tests for #905. This functionality is mentioned in the docs at http://lesscss.org/features/#import-atrules-feature.

The only issue with this change is once the `url()` support in #905 is added we end up adding an extra lookup since images & fonts don't end in `.less`. I could add an extra check for common image/font extensions if you'd like, or maybe move that to `plugin-css-url`.

### DIFF
--- a/e2e/fixtures/css/main.css
+++ b/e2e/fixtures/css/main.css
@@ -2,7 +2,4 @@
 @import 'import.css';
 
 /* @OctoLinkerResolve(<root>/css/import.css) */
-@import "import.css";
-
-/* @OctoLinkerResolve(<root>/css/import.css) */
-@import url("import.css") screen;
+@import url('import.css') screen;
--- a/e2e/fixtures/css/main.css
+++ b/e2e/fixtures/css/main.css
@@ -1,0 +1,8 @@
+/* @OctoLinkerResolve(<root>/css/import.css) */
+@import 'import.css';
+
+/* @OctoLinkerResolve(<root>/css/import.css) */
+@import "import.css";
+
+/* @OctoLinkerResolve(<root>/css/import.css) */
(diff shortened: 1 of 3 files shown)


e2e/fixtures/css/main.css
```diff
@@ -0,0 +1,8 @@
+/* @OctoLinkerResolve(<root>/css/import.css) */
+@import 'import.css';
+
+/* @OctoLinkerResolve(<root>/css/import.css) */
+@import "import.css";
```
stefanbuck, line 5: I think we can remove one of the test, since different quote marks are already covered in the RegExp test https://github.com/OctoLinker/OctoLinker/blob/main/packages/helper-grammar-regex-collection/test.js#L309-L310

e2e/fixtures/less/main.less
```diff
@@ -0,0 +1,11 @@
+/* @OctoLinkerResolve(<root>/less/import.less) */
+@import "import";
+
+/* @OctoLinkerResolve(<root>/less/import.less) */
+@import "import.less";
```
stefanbuck, line 5: Same as above

stefanbuck (approved):
Add one and a half comments, but apart from that this looks good

e2e/fixtures/css/main.css
xt0rted: I switched them all to single quotes since that's what we use elsewhere, and I removed the duplicate test.

stefanbuck:
Amazing, time for a new release ;)

PR #1035, opened 2020-09-24
Add support for scriptcs, Cake, and FAKE

xt0rted:
### Checklist:

Example files:

- [start.csx](https://github.com/scriptcs/scriptcs-samples/blob/master/wpf/start.csx)
- [HtmlAgilityPack.csx](https://github.com/amieres/FSharpStation/blob/e294fffadd7a9deeee9cd83b9fe3f4bf30607ac1/.paket/load/netstandard2.0/Test2/HtmlAgilityPack.csx)
- [build.cake](https://github.com/cake-build/cake/blob/develop/build.cake)
- [build.fsx](https://github.com/Shmew/Fable.SignalR/blob/master/build.fsx)
- [build.fsx](https://github.com/fsharp/FAKE/blob/release/next/build.fsx)


e2e/fixtures/dotnet/scripting/build.cake
```diff
@@ -0,0 +1,20 @@
+// @OctoLinkerResolve(https://www.nuget.org/packages/Cake.DotNetTool.Module)
+#module nuget:?package=Cake.DotNetTool.Module&version=0.4.0
+
+// @OctoLinkerResolve(https://www.nuget.org/packages/Cake.Coveralls)
+#addin "nuget:https://api.nuget.org/v3/index.json?package=Cake.Coveralls&version=0.10.1"
```
xt0rted, line 5: @stefanbuck I'm having trouble getting these lines to work as I wanted.

I was trying to make everything after `#module` & `#addin` the link so you didn't have to search for the package name in the url, but GitHub's parsing the lines with a full url and everything after `//` is in a span so no link gets injected.

Is there a way to support making that whole value a link or are we stuck with making just the package name the link?

This is the file I was testing against https://github.com/cake-build/cake/blob/develop/build.cake.

stefanbuck: This sounds similar to an issue I'm currently facing getting PHP / laravel support in. 

```html
<td>
  <span class="pl-k">use</span><span data-rgh-whitespace="space"> </span><span class="pl-v">Illuminate</span>\<span class="pl-v">Contracts</span>\<span class="pl-v">Container</span>\<span class="pl-v">Container</span>;
</td>
```

Yesterday, I spent two hours try to come up with a solution without luck. It's mainly a lack of support for such markup in https://github.com/padolsey/findAndReplaceDOMText which we use in the `insert-link` package. 

I've noticed that [refined-github](https://github.com/sindresorhus/refined-github) is able to wrap the url in your case. If we find a solution we should make sure it doesn't conflict with `refined-github` which is a widely used extension as well. 

Is this a blocker for you? Maybe best to ignore it for now and investigate into it separately.

xt0rted: No not a blocker. I can have just the package name be the link until we find a better solution.

stefanbuck:
My other PR was failing as well for some odd reason. After rebasing it passed

e2e/automated.test.js
```diff
@@ -43,14 +43,21 @@ describe('End to End tests', () => {
...
+          try {
+            await executeTest(
+              url,
+              targetUrl,
+              `#LC${lineNumber} .octolinker-link`,
+            );
+          } catch (ex) {
+            process.stdout.write(
+              `\n::error file=${file},line=${lineNumber}::${ex.message}\n`,
+            );
```
stefanbuck, line 58: 😍

stefanbuck: I think a `console.log` should work here too. That way you can get rid of the newline operator.

packages/plugin-dotnet-scripting/index.js
```diff
@@ -0,0 +1,31 @@
...
+import nugetResolver from '@octolinker/resolver-nuget';
+import relativeFile from '@octolinker/resolver-relative-file';
+
+export default {
+  name: 'DotNetScripting',
+
+  resolve(path, [target]) {
+    // this reference doesn't resolve to anything
+    if (target === './.fake/build.fsx/intellisense.fsx') {
+      return [];
```
stefanbuck, line 15: Just to share some insights, you can use `undefined` as a return value as well. Just saying, no need to change it.

packages/plugin-dotnet-scripting/index.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    return [relativeFile({ path, target }), nugetResolver({ target })];
```
stefanbuck, line 18: The signature of the `resolve` handler is `resolve(path, target, meta, regExp)`. 

Side note: Meta defaults to an empty object and is used in the context of [plugin-npm-manifest](https://github.com/OctoLinker/OctoLinker/blob/main/packages/plugin-npm-manifest/index.js#L19).

What I want to talk about is the fourth argument `regExp` which is the matching RegExp from line 29. In my PHP PR I make use of this to use different resolvers https://github.com/OctoLinker/OctoLinker/blob/87a137aa312f2a6e4b425f2e599744e2fa4f61f5/packages/plugin-php/index.js#L7-L13 maybe we can do something similar here. This allows use to be more specific and does reduce unnecessary look up in the OctoLinker API.

stefanbuck (approved):
Overall looks great. I left one comment about a potential optimization

packages/plugin-dotnet-scripting/index.js
xt0rted: That's awesome, I'll have to revisit some of the css & dotnet things I was working on because I'm pretty sure I'll be able to simplify them by checking the regex.

packages/plugin-dotnet-scripting/index.js
xt0rted: @stefanbuck I just tried `return;` and `return undefined;` and both give me an error. I guess that's why I was using an empty array here.

```
Uncaught (in promise) TypeError: Cannot read property 'registry' of undefined
    at app.js:41806
    at Array.map (<anonymous>)
    at app.js:41795
    at app.js:40329
    at Array.map (<anonymous>)
    at run (app.js:40322)
    at app.js:40355
    at gitHubInjection (app.js:15027)
    at init (app.js:40350)
```

The error points back to this line https://github.com/OctoLinker/OctoLinker/blob/dbaaaa5d845ed83198043201150753539210ee09/packages/helper-normalise-resolver-results/index.js#L67

stefanbuck: Ohh sorry for pointing you into the wrong direction. I guess there is some value in having a consistent return type 😉

xt0rted:
Once this is merged a PR needs to be sent to https://github.com/fsprojects/awesome-fsharp/issues/125 adding OctoLinker to the list.

PR #1036, opened 2020-09-24
Exclude urls from css imports

xt0rted:
[](https://github.com/OctoLinker/OctoLinker/pull/1036/checks) [Pull Request Badge](https://pullrequestbadge.com/?rel=badge)

### Checklist:

This is another thing I found while working on #905. I was trying to exclude things like `url("https://...")` so we didn't do needless lookups and it turns out the same thing happens with css imports. I'm sure this exists for LESS and SASS as well but didn't dig into those yet.

I'm wondering if there's a better way to handle this. Any ideas?

You can see an example of this here:
- https://github.com/meninaveneno/ROCK-STORY/blob/f0c8d332517c70161420da49feeaf7c786d7fb40/ROCKing.css

### DIFF
--- a/packages/helper-grammar-regex-collection/index.js
+++ b/packages/helper-grammar-regex-collection/index.js
@@ -140,7 +140,12 @@ export const CSS_IMPORT = regex`
   @import
   \s
   ((url|URL)\()?
-  ${captureQuotedWord}
+  ['"]                # beginning quote
+  (?<$1>
+    (?!https?:\/\/)   # exclude urls
+    [^'")]+           # capture the word inside the quotes
+  )
+  ['"]                # end quote
 `;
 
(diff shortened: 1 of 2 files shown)


packages/helper-grammar-regex-collection/index.js
```diff
@@ -140,7 +140,12 @@ export const CSS_IMPORT = regex`
   @import
   \s
   ((url|URL)\()?
-  ${captureQuotedWord}
+  ['"]                # beginning quote
+  (?<$1>
+    (?!https?:\/\/)   # exclude urls
```
stefanbuck, line 145: Using a negative look ahead, is what I ended up doing for #1033 as well.

stefanbuck (approved):

PR #1048, opened 2020-10-02
Add support for c++ project files, and C & C++ files

xt0rted:
### Checklist:

Example files:

- https://github.com/microsoft/cppwinrt/blob/ff7ab25319394eea2432b8d9ba8e50402ebee536/cppwinrt/cppwinrt.vcxproj
- https://github.com/microsoft/cppwinrt/blob/ff7ab25319394eea2432b8d9ba8e50402ebee536/cppwinrt/cppwinrt.vcxproj.filters

Working on this made me realize the `dot-net-project` plugin should probably be renamed to `msbuild` since all of the files it handles are msbuild files and not specifically .net related.


packages/plugin-c/index.js
```diff
@@ -0,0 +1,36 @@
...
+        /\.hpp$/,
+        /\.hxx$/,
+        /\.h%2B%2B.*$/, // .h++ needs to be url encoded to match
+      ],
+      githubClasses: [],
+    };
+  },
+
+  getLinkRegexes() {
+    return [C_INCLUDE];
```
stefanbuck, line 34: Do you mind adding an E2E test for this plugin as well. It doesn't have to be an E2E test for every single file extensions, maybe just a single one for `.cpp`? What do you think?

stefanbuck:
> Working on this made me realize the dot-net-project plugin should probably be renamed to msbuild since all of the files it handles are msbuild files and not specifically .net related.

I agree, although it's not a huge problem, since the name is not user facing and mainly just confusing for contributes (maybe)

packages/plugin-c/index.js
xt0rted: Yea, there's a few things missing from this one still. I'll get it finished up over the weekend.

e2e/fixtures/dotnet/c/project.vcxproj
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <ItemGroup>
+    <!-- @OctoLinkerResolve(<root>/dotnet/c/foo.cpp) -->
+    <ClCompile Include="foo.cpp" />
+
+    <!-- @OctoLinkerDoesNotResolve -->
+    <ClCompile Include="$(OutDir)strings.cpp">
```
xt0rted, line 9: @stefanbuck this test failed because it didn't resolve the dependency, and when that happens OctoLinker still injects an `<a>` but doesn't set the `href` attribute. How do we want to handle that scenario?

This is what the html looks like for this line:

```html
<td class="blob-code blob-code-addition">
  <span class="blob-code-inner blob-code-marker" data-code-marker="+">
    &lt;
    <span class="pl-ent">ClCompile</span>
    <span class="pl-e">Include</span>
    =
    <span class="pl-s">
      <span class="pl-pds">"</span>
      <a data-pjax="true" class="octolinker-link">$(OutDir)strings.cpp</a>
      <span class="pl-pds">"</span>
    </span>
    &gt;
  </span>
</td>
```

xt0rted: The false positives are resolved in #1069. Will rebase & finish up once that fix gets merged.

PR #1049, opened 2020-10-02
Remove jquery

xt0rted:
[](https://github.com/OctoLinker/OctoLinker/pull/1049/checks) [Pull Request Badge](https://pullrequestbadge.com/?rel=badge)

### Checklist:

I'm not sure if this is complete, it's been a few months since I worked on it. My hope was to remove a dependency we didn't need anymore since it's only used in this one file. Is this change something the tests should sufficiently cover? I don't see any errors while browsing around but it's possible something is silently broken.

### DIFF
--- a/packages/blob-reader/helper.js
+++ b/packages/blob-reader/helper.js
@@ -1,5 +1,3 @@
-import $ from 'jquery';
-
 function getBlobCodeInner(el) {
   return [].slice.call(el.getElementsByClassName('blob-code-inner'));
 }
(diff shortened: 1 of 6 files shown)


packages/blob-reader/helper.js
```diff
@@ -117,24 +122,26 @@ function getLineNumber(el) {
   }
 
   // split diff view
-  let lineNumber = $(el).closest('td').prev().data('line-number');
+  let lineNumber = Number(
+    el.closest('td').previousElementSibling.dataset.lineNumber,
+  );
```
stefanbuck, line 127: To increase false tolerance we might want to use `parseInt` over `Number`. 

```js
parseInt("123abc", 10)
// returns 123

Number("123abc")
// returns NaN
```

packages/blob-reader/helper.js
```diff
@@ -63,40 +61,47 @@ function getPath(el) {
...
-  let ret = $(rootSelector)
-    .filter(function () {
-      return $(this).text().trim() === 'View file';
-    })
-    .attr('href');
+  const [link] = Array.from(rootSelector).filter((value) => {
+    return value.textContent.trim() === 'View file';
+  });
+
+  let ret = link && link.getAttribute('href');
```
stefanbuck, line 68: There is nothing wrong with this line, but optional chaining might be a fit here 

```suggestion
  let ret = link?.getAttribute('href');
```

packages/blob-reader/helper.js
```diff
@@ -63,40 +61,47 @@ function getPath(el) {
...
 
   if (!ret) {
-    ret = $('.js-permalink-shortcut').attr('href');
+    const shortcut = document.querySelector('.js-permalink-shortcut');
+
+    ret = shortcut && shortcut.getAttribute('href');
```
stefanbuck, line 73: Again, I think optional chaining helps to make this more readable.

```suggestion
    ret = document.querySelector('.js-permalink-shortcut')?.getAttribute('href');
```

packages/blob-reader/helper.js
```diff
@@ -63,40 +61,47 @@ function getPath(el) {
...
-    .attr('href');
+  const [link] = Array.from(rootSelector).filter((value) => {
+    return value.textContent.trim() === 'View file';
+  });
+
+  let ret = link && link.getAttribute('href');
 
   if (!ret) {
-    ret = $('.js-permalink-shortcut').attr('href');
+    const shortcut = document.querySelector('.js-permalink-shortcut');
```
stefanbuck, line 71:
```suggestion

```

packages/blob-reader/helper.js
```diff
@@ -63,40 +61,47 @@ function getPath(el) {
...
+
+    ret = shortcut && shortcut.getAttribute('href');
   }
 
   // When current page is a gist, get path from blob name
   if (!ret && isGist()) {
-    ret = $('.gist-blob-name', el.parentElement).text().trim();
+    ret = el.parentElement.querySelector('.gist-blob-name').textContent;
```
stefanbuck, line 78: The nice thing about jQuery is the build in fault tolerance. If the selector is not matching anything the result of the operation on the left was `undefined` whereas on the right, it will throw an error `Cannot read property 'textContent' of null`

Again, optional chaining is here to help

```suggestion
    ret = el.parentElement.querySelector('.gist-blob-name')?.textContent;
```

stefanbuck: Do you mind adding optional chaining everywhere?

stefanbuck (changes requested):
Overall this pull request looks great, THANK YOU!

I just left a few comments about the usage of optional chaining and other minor things. However, I would prefer to hold this PR back until the next release is out so we get more time to test it.

packages/blob-reader/helper.js
xt0rted: I keep forgetting this is supported in JS now.

packages/blob-reader/helper.js
xt0rted: Nope, I prefer it. I use it in TypeScript but keep forgetting it's supported in JS now too.

packages/blob-reader/helper.js
xt0rted: @stefanbuck I tried using optional chaining in another PR and Webpack got angry and gave a bunch of errors. Turns out support for them wasn't added until [Acorn 7.3.0](https://github.com/acornjs/acorn/releases/tag/7.3.0) but [Webpack v4 uses Acorn v6.x](https://github.com/webpack/webpack/blob/2efeb4b5787e868e69233aa9dec469c200fa3b2c/package.json#L12) 😞 

According to https://github.com/webpack/webpack/issues/10227#issuecomment-653408484 there's a way to get it to work in Webpack v4 when using Yarn but I haven't tried setting it up yet.

xt0rted:
Closing in favor of #1420.

PR #1050, opened 2020-10-02
Ensure release note is updated

stefanbuck:
[](https://github.com/OctoLinker/OctoLinker/pull/1050/checks) [Pull Request Badge](https://pullrequestbadge.com/?rel=badge)

This change ensures that a release description is added before performing a release using `npm version <patch|minor|major>`

### DIFF
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+# This script will test if you have given a leap year or not.
+
+echo "Please summarize this release in one sentence"
+
+read note
+
+dot-json assets/manifest.json version $npm_package_version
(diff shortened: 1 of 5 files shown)


scripts/version.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# This script will test if you have given a leap year or not.
```
xt0rted, line 2: I don't think this is related to what this is doing.

xt0rted (approved):
The script comment looks like it might be wrong, otherwise looks good to me.

scripts/version.sh
stefanbuck: haha copy & 🍝

PR #1063, opened 2020-10-06
Creating annotations for failed e2e tests

xt0rted:
[](https://github.com/OctoLinker/OctoLinker/pull/1063/checks) [Pull Request Badge](https://pullrequestbadge.com/?rel=badge)

### Checklist:

I split this out of #1035 since it's not related to that piece of work.

I'm wondering if we should take a dependency on [actions/core](https://github.com/actions/toolkit/tree/main/packages/core) so we can use their [`issueCommand`](https://github.com/actions/toolkit/blob/main/packages/core/src/command.ts) function which handles building & escaping everything.

### DIFF
--- a/e2e/automated.test.js
+++ b/e2e/automated.test.js
@@ -16,6 +16,11 @@ async function executeTest(url, targetUrl, selector) {
   await page.waitForSelector(`${selector}[href$="${targetUrl}"]`);
 }
 
+function createAnnotation({ file, lineNumber, ex }) {
+  // eslint-disable-next-line no-console
+  console.log(`::error file=${file},line=${lineNumber}::${ex.message}`);
+}
+
 describe('End to End tests', () => {
   beforeAll(async () => {
     if (!process.env.E2E_USER_NAME || !process.env.E2E_USER_PASSWORD) {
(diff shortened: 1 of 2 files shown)


stefanbuck:
> I'm wondering if we should take a dependency on [actions/core](https://github.com/actions/toolkit/tree/main/packages/core) so we can use their [`issueCommand`](https://github.com/actions/toolkit/blob/main/packages/core/src/command.ts) function which handles building & escaping everything.

I like the simplicity of the `console.log`, but you made a good point. I think this is something we can look into in separate PR. This is already a huge improvement. Thanks for looking into this.

e2e/automated.test.js
```diff
@@ -16,6 +16,11 @@ async function executeTest(url, targetUrl, selector) {
   await page.waitForSelector(`${selector}[href$="${targetUrl}"]`);
 }
 
+function createAnnotation({ file, lineNumber, ex }) {
+  // eslint-disable-next-line no-console
+  console.log(`::error file=${file},line=${lineNumber}::${ex.message}`);
```
staabm, line 21: To support multi-line messages you need to encode `\n` with `%0A`

See https://github.com/mheap/phpunit-github-actions-printer/pull/14

xt0rted:
@stefanbuck looks like switching to `console.log()` doesn't work as expected. Some messages get written out, but others don't because of how the tests are async but not everything is awaited.

https://github.com/OctoLinker/OctoLinker/pull/1048/checks#step:10:51

I tried making everything async but the callback to `describe()` can't be async so I'm back to square one now.

PR #1066, opened 2020-10-09
Add laravel support

stefanbuck:
Add feature preview to support to link `use` statements from the following projects:

- Symfony
- Laravel
- Doctrine
- CakePHP
- PHPUnit

To enable the feature, go to the extension settings page and enable the `Better PHP support` checkbox

### DIFF
--- a/e2e/__snapshots__/automated.test.js.snap
+++ b/e2e/__snapshots__/automated.test.js.snap
@@ -1,18 +1,18 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`End to End tests diff view links commentbox 1`] = `27`;
+exports[`End to End tests diff view links commentbox 1`] = `28`;
 
 exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/commit/b97dfbfdbf3dee5f4836426e6dac6d6f473461db?diff=split#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L23 to https://github.com/eslint/eslint 1`] = `27`;
 
 exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/commit/b97dfbfdbf3dee5f4836426e6dac6d6f473461db?diff=split#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R24 to https://github.com/eslint/eslint 1`] = `27`;
 
-exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fL2 to https://github.com/webpack-contrib/copy-webpack-plugin 1`] = `27`;
+exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fL2 to https://github.com/webpack-contrib/copy-webpack-plugin 1`] = `28`;
 
-exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fR2 to https://github.com/webpack-contrib/copy-webpack-plugin 1`] = `27`;
+exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fR2 to https://github.com/webpack-contrib/copy-webpack-plugin 1`] = `28`;
 
-exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L42 to https://github.com/eslint/eslint 1`] = `27`;
+exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L42 to https://github.com/eslint/eslint 1`] = `28`;
 
-exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R42 to https://github.com/eslint/eslint 1`] = `27`;
+exports[`End to End tests diff view resolves https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R42 to https://github.com/eslint/eslint 1`] = `28`;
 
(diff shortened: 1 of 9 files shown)


packages/helper-insert-link/index.js
```diff
@@ -43,6 +43,21 @@ function injectUrl(node, value, startOffset, endOffset) {
...
+      if (textMatch.includes(child.textContent)) {
+        if (!el.childElementCount) {
+          node.appendChild(el);
+        }
+        el.appendChild(child);
+      } else {
+        node.appendChild(child);
+      }
+    });
+  }
```
stefanbuck, line 59: This might be all we need to get rid of `findAndReplaceDOMText`, let's see ....

packages/plugin-php/index.js
```diff
@@ -5,8 +5,18 @@ export default {
   name: 'PHP',
 
   resolve(path, [target], meta, regExp) {
+    if (target.startsWith('Illuminate')) {
+      return liveResolverQuery({
+        type: 'ping',
+        target: `https://laravel.com/api/8.x/${target.replace(
+          /\\/g,
+          '/',
+        )}.html`,
```
xt0rted, line 14: Should we move this out to a helper function or maybe move it down lower in the api? It's being done over here too https://github.com/OctoLinker/OctoLinker/blob/4317c15952d159ef6d503b7f8b887666b86fc8c7/packages/plugin-dotnet-project/index.js#L8-L9

packages/helper-insert-link/index.js
xt0rted: Just tested this on #1035 and I'm able to turn the whole selection into a link now 🎉

image

stefanbuck: Awesome! 

In the current version of the `insert-link` package we strip out the quotation marks, to link the actual text without the surrounding quotation marks. This is mainly driven from my own desire to get a nice looking link. This seems to be the root cause for the issue you were facing. However, this change needs further testing and therefore it might be best to split this out into a separate PR

packages/plugin-php/index.js
stefanbuck: This was just a quick try to test if this is working. I'm currently consider to add this logic to the OctoLinker API so we can maintain this more easily. Also I would like to avoid such a mapping list to make it work for all api docs out of the box which might be tricky or even impossible.

e2e/__snapshots__/automated.test.js.snap
```diff
@@ -1,18 +1,18 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`End to End tests diff view links commentbox 1`] = `27`;
+exports[`End to End tests diff view links commentbox 1`] = `28`;
```
stefanbuck, line 3: This is driving me nuts.

If I run `document.querySelectorAll('.octolinker-link').length` on https://github.com/OctoLinker/OctoLinker/pull/451/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166f, it return 29 on this branch as well as with the latest main build. 

I have no idea why this was 27 in the first place.

e2e/__snapshots__/automated.test.js.snap
```diff
@@ -266,12 +266,16 @@ exports[`End to End tests single blob resolves requireNamespace("rlang") to http
 
 exports[`End to End tests single blob resolves requireNamespace("rlang", quietly = TRUE) to https://github.com/r-lib/rlang 1`] = `7`;
 
-exports[`End to End tests single blob resolves use ArrayAccess; to https://www.php.net/manual/en/class.arrayaccess.php 1`] = `5`;
+exports[`End to End tests single blob resolves use ArrayAccess; to https://www.php.net/manual/en/class.arrayaccess.php 1`] = `7`;
```
stefanbuck, line 269: This is expected, because this PR adds support for laravel improts

stefanbuck:
[Dev build](https://github.com/OctoLinker/OctoLinker/files/5533551/OctoLinker-php.zip)

staabm:
tested in chrome 87.0.4280.63 on windows 10x64

works as expected on the following popular symfony based apps
- https://github.com/deployphp/deployer/blob/master/src/Deployer.php
- https://github.com/composer/composer/blob/master/src/Composer/Command/ArchiveCommand.php
- https://github.com/composer/composer/blob/master/src/Composer/Console/Application.php

within the symfony repo itself its a bit confusing, because we get octolink links as well as github-native linking (I guess it depends on timing, but sometimes the click on such a link is served by octolinker and sometimes by the github native linking.. I feel ocotolink should disable this feature when on the supported framework repos)
- https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/VarExporter/Internal/Hydrator.php

grafik

-------

while using the extension I found some other candidate frameworks/libs, which are pretty popular in the php ecosystem and might be usefull to support. besides the application frameworks you already support, you would also get some great coverage by supporting popular testing tools/libs like
- https://github.com/sebastianbergmann/phpunit
- https://github.com/nikic/PHP-Parser
- https://github.com/guzzle/guzzle

its a different angle on the same problem/solution and would open the foor for a lot of tools which are not essentially based on wone of the supported app-frameworks

--------

another thought.. while browsing in https://github.com/octobercms/october/blob/develop/bootstrap/app.php I thought, that octolinker could also link full qualified class names like 

https://github.com/octobercms/october/blob/5d0988559bb15cb93b034e100b0a246efca707f8/bootstrap/app.php#L29-L42

stefanbuck:
Thanks @staabm for detailed feedback, really appreciated.

I'm aware of the clash with GitHub's native code navigation when using OctoLinker. I found myself in a similar situation a few times. However at least for me the code navigation provided by OctoLinker was suiting my needs. Also, I did not received any other complains this regards, but this doesn't mean there is no issue. Anyway, I agree that in this particular case it would be better to disable OctoLinker. I need to investigate if this is feasible.

I'll take your suggestions about additional frameworks/libs into account. As discussed offline, adding coverage for other frameworks and libs should be easy and I want to work on improving this experience as well. The current implementing is quite verbose and needs another thought.

Adding support for  full qualified classes should be simple, but I wonder how common this is. Happy to discuss this further.

Regarding the editor integration. This was requested long time ago already https://github.com/OctoLinker/OctoLinker/issues/280. I can see how useful such an extension would be, and to be honest I'm seeking for it myself. However, creating an OctoLinker like extension is not a trivial task. There is a fundamental difference how plugins for VSCode work compared to OctoLinker in the browser. OctoLinker is using regular expression to find the snippets whereas in a plugin this needs to be done via an AST. And there is also the issue with the user expectation. OcotLinker supports a bunch of different languages and rebuilding all of this in an editor plugin sounds like a big task. I'm more than happy to move this forward if someone wants to take the lead.

stefanbuck:
I'm going to release this soon behind a preview flag. This allows us to collect feedback and make changes as we go along.

image

packages/helper-settings/index.js
```diff
@@ -9,13 +9,35 @@ const defaults = {
     since: Date.now(),
     counter: 0,
   },
+  enableBetterPHP: false,
 };
 
+function isTestEnv() {
+  if (process.env.NODE_ENV === 'test') {
+    return true;
+  }
```
stefanbuck, line 18: This is needed to make unit test work for functionality behind a feature flag

packages/helper-settings/index.js
```diff
@@ -9,13 +9,35 @@ const defaults = {
...
+
+  if (
+    typeof window !== 'undefined' &&
+    /OctoLinker\/OctoLinker\/blob\/.+\/e2e\/fixtures/.test(
+      window.location.pathname,
+    )
+  ) {
+    return true;
```
stefanbuck, line 26: This is needed to make e2e test work for functionality behind a feature flag

stefanbuck:
@xt0rted this is finally ready for review.

PR #1067, opened 2020-10-09
Fix return value of PHP plugin

stefanbuck:
The return value from a resolve needs to be an array as discussed here https://github.com/OctoLinker/OctoLinker/pull/1035#discussion_r500038960. This might be worth a patch release


xt0rted (approved):

stefanbuck:
So I had a look with fresh eyes and the return value can be `undefined` in this case, because the `insert-link` package is unable to wrap the given match due to the issue fixed in #1067. Anyway, let's get this change in but without a patch release since it does not throw any exception.

PR #1072, opened 2020-10-10
Update package name from JSONPath to jsonpath-plus and update to the latest version

xt0rted:
### Checklist:

This package [changed it's name](https://github.com/JSONPath-Plus/JSONPath/blob/master/CHANGES.md#0120-dec-12-2015-1039pm) years ago which is why we haven't received any updates for it. There shouldn't be any changes between these two versions exception for the name change. Should I try updating to a newer version or should be a different PR?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
     - uses: actions/setup-node@v1
       with:
         node-version: '12.x'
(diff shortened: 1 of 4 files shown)


stefanbuck:
Good catch. Yes, feel free to do the update as well.

stefanbuck (approved):

packages/helper-process-json/index.js
```diff
@@ -1,4 +1,4 @@
-import jsonPath from 'JSONPath';
+import { JSONPath as jsonPath } from 'jsonpath-plus';
```
xt0rted, line 1: I did this because eslint complains about `JSONPath` being capitalized and not being used as a constructor.

.github/workflows/ci.yml
```diff
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
```
xt0rted, line 13: @stefanbuck looks like this fixes the test issue we've been having. Not sure why it didn't do anything in #1035 but it worked here.

PR #1084, opened 2020-10-12
Rstats library

edavidaja:
I'm still puzzling through the regex handling for R manifests, which may make sense to handle via a separate plugin, as with Ruby or Rust, but this covers scripts and RMarkdown files.

### Checklist:

### DIFF
--- a/e2e/fixtures/r/foo.R
+++ b/e2e/fixtures/r/foo.R
@@ -1,0 +1,20 @@
+# @octoLinkerResolve(https://github.com/r-lib/rlang)
+library(rlang)
+
+# @octoLinkerResolve(https://github.com/r-lib/rlang)
+library("rlang")
+
+# @octoLinkerResolve(https://github.com/r-lib/rlang)
+require(rlang, quietly = TRUE)
+
+# @octoLinkerResolve(https://github.com/r-lib/rlang)
+require("rlang")
+
+# @octoLinkerResolve(https://github.com/r-lib/rlang)
+requireNamespace("rlang", quietly = TRUE)
+
+# @octoLinkerResolve(https://github.com/r-lib/rlang)
+requireNamespace("rlang")
+
+# @octoLinkerResolve(https://github.com/r-lib/rlang)
+loadNamespace("rlang")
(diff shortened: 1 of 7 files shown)


stefanbuck:
Tests are failing, because we download the markup from https://github.com/OctoLinker/testrepo/issues/2 which was missing a R-snippet (I just added that). If you run `sh ./packages/blob-reader/scripts/update-fixtures.sh` it will download all fixtures file. Please reject all changes expect `packages/blob-reader/fixtures/github.com/issue/code.html`. This should make the test pass

stefanbuck (commented):
Thanks for taking the time to add R support to OctoLinker. This looks really good already. Let me know once you're done with all the regex stuff. Regarding the manifest file, I agree this should be a different plugin.

edavidaja:
It looks like the e2e tests are timing out? Apologies for blowing up your inbox--I should have run some of those tests locally first.

e2e/fixtures/r/foo.R
```diff
@@ -0,0 +1,20 @@
+# @octoLinkerResolve(https://github.com/r-lib/rlang)
```
xt0rted, line 1: @stefanbuck are these case sensitive?

stefanbuck: Yes, good spot

stefanbuck:
You might need to rebase your branch to fix the E2E test timeout. 

@xt0rted the branch is 10 commits behind OctoLinker's `main` branch which seems to be the issue as discussed a few days ago https://github.com/OctoLinker/OctoLinker/pull/1084/checks?check_run_id=1244296101#step:10:203

e2e/fixtures/r/foo.R
stefanbuck: @edavidaja please update all OctoLinker annotations to ensure E2E tests are running

```diff
- @OctoLinkerResolve
+ @OctoLinkerResolve
```

xt0rted:
@stefanbuck take a look at https://github.com/OctoLinker/OctoLinker/pull/1072#discussion_r502846041. That change should fix this issue, it worked on that PR.

e2e/fixtures/r/foo.R
edavidaja: fixed

PR #1120, opened 2020-11-11
Adjust toast position for Octotree

stefanbuck:
As outlined in https://github.com/ovity/octotree/issues/997, Octotree is hiding the toast notification that we introduced recently. Usually I would like to avoid such tweaks, but since Octotree is a widely used extension it's worth a fix in my opinion. 

## Before
Screenshot 2020-11-11 at 22 06 31

## After
Screenshot 2020-11-11 at 22 06 08

### DIFF
--- a/packages/user-interface/style.css
+++ b/packages/user-interface/style.css
@@ -2,6 +2,10 @@
   pointer-events: none;
 }
 
+.octotree-show .octolinker-toast  {
+  left: var(--octotree-sidebar-width) !important
+}
+
 .octolinker-toast .octo-toast-image  {
   width: 58px;
   height: 40px;


packages/user-interface/style.css
```diff
@@ -2,6 +2,10 @@
   pointer-events: none;
 }
 
+.octotree-show .octolinker-toast  {
+  left: var(--octotree-sidebar-width) !important
```
stefanbuck, line 6: It feels wrong that we have to really on class names and variables added by another browser extension, but this is the best solution I was able to come up with. I tried to move the toast notification into another diff so it gets pushed with the entire content, but this didn't worked out.

stefanbuck:
FYI @jdreesen

packages/user-interface/style.css
xt0rted: Adjusting the z index wasn't enough?

stefanbuck: Do be honest, I haven't really thought about that 😆  I can give it another try this weekend. Thanks for the hint

PR #1122, opened 2020-11-14
Prevent chrome from crashing

### LINKED ISSUE #1121: Chrome GitHub pages crashing when OctoLinker enabled
**Describe the bug**
When loading certain GitHub pages in Chrome, OctoLinker causes the page to go unresponsive after partially loading. Page content is partially or fully rendered, but content cannot be interacted with (text selected, etc.) and switching to another tab and back displays a blank page.

**To Reproduce**
1. Disable all extensions
2. Visit https://github.com/framer/motion/issues/364 (the first page where I noticed this happening)
3. Page loads and is responsive
4. Enable only OctoLinker extension
5. Reload the page or open it in a new tab
6. Page never finishes loading

**Expected behavior**
The page should fully load in a timely manner while retaining expected interactivity.

**Additional context**

I noticed that when I load the page, two network requests are fired that resolve almost immediately:

image

After enabling OctoLinker, the two network requests occur but never leave the "Pending" status:

image

I allowed the tab to sit for ~15 minutes and the requests never completed. 

OctoLinker version: 6.6.0
Chrome version: Tested with 86.0.4240.193 and 86.0.4240.198
OS: macOS 10.15.5

stefanbuck:
Fix #1121. When viewing issue comments with snippets attached that contains relative imports OctoLinker was causing a lot of CPU consumption. The root cause was the regular expression we used to figure out whether or not a url is valid or not. Code snippets doesn't have a root path (because they are not files) which results in constructed urls like `https://github.comframer-motion-supported-elements/index.js`. Switching to another regular expression seems to fix this issue. In addition I added an early return within the `javascriptFile` resolver if such a root path is not given.


packages/plugin-javascript/__tests__/__snapshots__/index.js.snap
```diff
@@ -77,36 +77,36 @@ Array [
...
-  "{BASE_URL}modules/es6.symbol.ts",
-  "{BASE_URL}modules/es6.symbol/index.ts",
-  "{BASE_URL}modules/es6.symbol.tsx",
-  "{BASE_URL}modules/es6.symbol/index.tsx",
-  "{BASE_URL}modules/es6.symbol.ls",
-  "{BASE_URL}modules/es6.symbol/index.ls",
-  "{BASE_URL}modules/es6.symbol.json",
-  "{BASE_URL}modules/es6.symbol/index.json",
-  "{BASE_URL}modules/es6.symbol",
+  "{BASE_URL}/blob/path/modules/es6.symbol.js",
```
stefanbuck, line 80: Expected change, since the original test wasn't passing a root path which is now required to make this work

PR #1123, opened 2020-11-15
Adjust toast position for Octotree

stefanbuck:
As suggested in https://github.com/OctoLinker/OctoLinker/pull/1120#discussion_r522915677 this PR is using `z-index` to ensure the OctoLinker toast message stays on top

## After
Screenshot 2020-11-15 at 22 25 42

## Before
Screenshot 2020-11-15 at 22 25 20

### DIFF
--- a/packages/user-interface/style.css
+++ b/packages/user-interface/style.css
@@ -1,3 +1,6 @@
+.octolinker-toast {
+  z-index: 1000000002 !important;
+}
 .octolinker-toast svg {
   pointer-events: none;
 }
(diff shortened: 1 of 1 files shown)


packages/user-interface/style.css
```diff
@@ -1,11 +1,10 @@
+.octolinker-toast {
+  z-index: 1000000002 !important;
```
stefanbuck, line 2: Octotree is using `1000000001` as `z-index` value

jdreesen:
This works great, thx for the fix! :heart:

PR #1129, opened 2020-11-18
Show usage statistics on setttings page

stefanbuck:
Since https://github.com/OctoLinker/OctoLinker/pull/917 we track the usage of OctoLinker by increasing a counter whenever a user follows a link. The data are stored in the users browser using the Storage API. This PR makes this statiscs accesibale by adding a new section to the settings page.

Reviewer note: [ignore whitespace in diff using this link](https://github.com/OctoLinker/OctoLinker/pull/1129/files?w=1)

### DIFF
--- a/packages/helper-settings/components/stats.js
+++ b/packages/helper-settings/components/stats.js
@@ -1,0 +1,8 @@
+import { h } from 'preact';
+
+export default ({ counter }) => (
+  <div>
+    <h2>Statistics</h2>
+    Links followed: <strong>{counter}</strong>
+  </div>
+);
--- a/packages/helper-settings/components/stats.js
+++ b/packages/helper-settings/components/stats.js
@@ -4,5 +4,9 @@ export default ({ counter }) => (
   <div>
     <h2>Statistics</h2>
     Links followed: <strong>{counter}</strong>
+    <span className="note pt-4">
+      Your statistics are stored in your browser, we do not store them anywhere
+      else.
+    </span>
   </div>
 );
--- a/packages/core/stats.js
+++ b/packages/core/stats.js
@@ -5,10 +5,7 @@ const CLASS_NAME = 'octolinker-link';
 let tid = null;
 
 function updateCounter() {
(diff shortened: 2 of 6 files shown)


packages/helper-settings/components/stats.js
```diff
@@ -0,0 +1,14 @@
+import { h } from 'preact';
+
+const options = {
+  year: 'numeric',
+  month: 'long',
+  day: 'numeric',
+};
```
xt0rted, line 7: Is any of this being used?

stefanbuck: Sorry, I should have paid more addention to the diff before asking for review. To answer your quesiton, no it's not. I was thinking of showing the date of since clicks are tracked, but to keep the wording simple I removed it.

xt0rted: What's `h` used for?

stefanbuck: This is needed to make JSX work. JSX is not part of the JavaScript spec so we have to manual import it. I think there are also ways to automate this using a webpack plugin, but since the usage of preact is quite low it might not worth the effort.

stefanbuck:
@xt0rted are you happy with me merging this?

xt0rted (approved):
@stefanbuck I'm noticing some weird behavior with the settings page in chrome. When I open it on a fresh install nothing shows up:

image

Once I hover over an OctoLinker link then the settings page shows correctly:

image

It's also changing size depending on my screen size, but this may be intentional:

image

Beside this it looks good.

staabm:
The statistic should have a note, that no statistical data is stored outside the browser and everything is stored locally

stefanbuck:
Good shout @staabm

stefanbuck:
@staabm I added a note saying: 
> Your statistics are stored in your browser, we do not store them anywhere else.

stefanbuck:
@xt0rted the issue you’re facing is related to fresh installations. Without using the extensions the `stats` object remains empty. I made it part of the default settings which seems to fix the issue

PR #1147, opened 2020-12-11
Add python e2 test cases.

laundmo:
### Checklist:

This PR adds test cases for python imports. It's mostly focused on relative imports but also addresses an issue with importing from a library's subpackage.


e2e/fixtures/python/relative/subpackage2/import_from_other.py
```diff
@@ -0,0 +1,12 @@
+# @OctoLinkerResolve(<root>/python/relative/subpackage1/classes.py)
+from ..subpackage1.classes import MyClass1
+
+# @OctoLinkerResolve(<root>/python/relative/subpackage1/classes.py)
+from ...relative.subpackage1 import classes
+#    ^ should resolve folder        ^ should resolve file
```
laundmo, line 6: i dont know what to do here, i commented on this one case but there are similar cases everywhere

xt0rted: Without making any changes to the current test code you could try using both `@OctoLinkerResolve()` and `@OctoLinkerResolveAbove()` so the first comment would check for the first link on the line below it, and the second comment will check for the second link on the line above it.

An example of this is:

```python
# @OctoLinkerResolve(<root>/python/relative/subpackage1)
from ...relative.subpackage1 import classes
# @OctoLinkerResolveAbove(<root>/python/relative/subpackage1/classes.py)
```

If there's more than two links per line then this wouldn't work. If we do need to expand the test api then maybe something like this could work?

```python
# @OctoLinkerResolveList(foo.py;bar.py;baz.py)
from ...relative.subpackage1 import classes
```

laundmo: Thanks for this. Ideally OctoLinker (and the tests) should resolve each segment of the entire subpackage path (and allow for that to be shown in the tests) but for now this works.

stefanbuck:
Thanks for submitting the fixtures, I really appreciate your time working on this. I started looking into this and it seems like we need changes made in #1066 regarding the `helper-insert-linker` package to make some of the imports working. Let me finish that PR first and then I'll come back to this one.

e2e/fixtures/python/top_level_import.py
```diff
@@ -0,0 +1,22 @@
...
+# @OctoLinkerResolveAbove(<root>/python/relative/)
+
+# @OctoLinkerResolve(<root>/python/relative/simple.py)
+from relative import simple
+# @OctoLinkerResolveAbove(<root>/python/relative/)
+
+# @OctoLinkerResolve(<root>/python/relative/simple.py)
+from relative.simple import my_func
+
+# @OctoLinkerResolve(<root>/python/relative/subpackage1/classes)
```
stefanbuck, line 12: Shouldn't be this with `.py`?
```suggestion
# @OctoLinkerResolve(<root>/python/relative/subpackage1/classes.py)
```

stefanbuck:
In c8920f1 I disabled quite a few tests which are not supported yet. I want to get this PR merged asap. In a later PR I'll address the actual issue. Sorry that this PR was on hold for such a long time.

laundmo:
@stefanbuck no problem, i kinda assumed that this wasn't going to be merged until the features are implemented. still nice to see progress.

PR #1175, opened 2021-01-10
Fix Safari CI build

fregante:
Follows: https://github.com/OctoLinker/OctoLinker/pull/1167#issuecomment-757384115

From the log:
```
Artifact artifact was downloaded to /Users/runner/work/OctoLinker/OctoLinker/dist/artifact
9
```


stefanbuck:
Amazing thanks. I downloaded the artifact but didn't spotted the wrong name.

PR #1181, opened 2021-01-16
Safari release

### LINKED ISSUE #29: Safari support

fregante:
Closes #29 

I submitted the app for reviewing 🥳 

You can provide additional screenshots or videos:

- images: one of  1280 x 800, 1440 x 900, 2560 x 1600, 2880 x 1800
- videos: 1080p, https://user-images.githubusercontent.com/1402241/104794582-2ee96980-576e-11eb-8a74-005b59832ba7.png


safari/app-store-description.txt
```diff
@@ -0,0 +1,11 @@
...
+
+OctoLinker reads the source code you're currently viewing straight from the browser or, in some cases, using the GitHub API.
+
+No source code will ever be sent to a server! Your source code stays in your browser.
+
+Based on the file type, OctoLinker invokes a plugin which then executes a regular expression on the source code to find import statements.
+
+Relative file references are resolved through the GitHub tree API in your browser. External dependencies are resolved using the OctoLinker API.
+
+Only the name of a dependency along with the registry type is sent to this service.
```
fregante, line 11: I suggest having a single copy-pastable description that every store uses because they all have different copy now, with Chrome’s copy including some duplicate/non-clickable links.

I copied this text from the website. 

File named like https://github.com/sindresorhus/refined-github/commit/d185f27dabd85aa4b7c3f61a3f9d5ca9c6c40c1d#diff-4810e2862897d3e19bd39b745d6adae0c80afac122abf4e2aa2b845a9649fd5a

fregante:
1. Install from app store
2. Go to Safari’s Preferences
3. Extensions tab
4. Enable the extension
5. Click "Edit websites"
    gif
6. Try it out at https://github.com/OctoLinker/OctoLinker/blob/main/package.json
  Screen Shot 13

fregante:
I don't expect to update every single version manually because it's a pain, so after this version we'll probably wait for https://github.com/sindresorhus/refined-github/issues/3686

safari/app-store-description.txt
stefanbuck, line 11:
```suggestion
How does it work?

OctoLinker reads the source code you're currently viewing straight from the browser. 

No source code will ever be sent to a server! Your source code stays in your browser. 

Based on the file type, OctoLinker invokes a plugin which then executes a regular expression on the source code to find import statements.

Relative file references are resolved through the GitHub tree API in your browser. External dependencies are resolved using the OctoLinker API (http://github.com/octoLinker/api).

Only the name of a dependency along with the registry type is sent to this service.
```

stefanbuck: Totally agree. I made some suggestions and once we agreed on the copy I'll update all the other store listings.

stefanbuck:
> I don't expect to update every single version manually because it's a pain, so after this version we'll probably wait for [sindresorhus/refined-github#3686](https://github.com/sindresorhus/refined-github/issues/3686)

Fair enough. I'll let you know if there is manual update needed in case of a serious issue but other than that I agree we should wait for the auto upload.

stefanbuck:
> * Install from the [app store](https://apps.apple.com/app/octolinker/id1549308269)
> * Go to Safari’s Preferences
> * Extensions tab
> * Enable the extension
> * Click "Edit websites"

I wonder if we should provide those steps after the extension is installed ... or is this somewhat clear as a Safari user? We could create a safari setup page on https://octolinker.now.sh/ and may redirect in case of a Safari installation.

fregante:
> I wonder if we should provide those steps after the extension is installed

I wonder too, but perhaps the user is already nudged towards the solution:

1. After downloading from the app store, the button becomes "open" (or they can see the app in the launcher)
2. When the app is opened, the app tells that the extension is disabled and has a button to open the preferences on the exact tab
3. On that tab they can see that the extension is disabled and can be enable.
4. When enabling the app, they should see a new toolbar button
5. Clicking the button on GitHub.com triggers this message: 

	

And really, the only step that we can provide is the very last one because that's the only time we can run code in an extension. Honestly I wouldn't mind having a document that explains every step just in case the user gets lost, but I don't think many would see it.

stefanbuck:
> Honestly I wouldn't mind having a document that explains every step just in case the user gets lost, but I don't think many would see it.

Alright, let's see if this needs further explanation before we spend time on this.

fregante:
Feel free to merge at any time. Unfortunately descriptions and screenshots are part of the same version review on the app store, so they can only be updated when the next version is uploaded.

stefanbuck:
@fregante I commited all the store assets (5 images + 1 video) https://github.com/OctoLinker/media/commit/785f70f395803f38b09743606c9f3714f26659a1

fregante:
The video has to be exactly 1920x1080 though, or else it's not accepted

fregante:
Same for the photos. The resolution should be 1280x900, not x800

safari/app-store-description.txt
```diff
@@ -0,0 +1,13 @@
+OctoLinker is the easiest and best way to navigate between files and projects on GitHub. It turns language-specific statements like include, require or import into links. It supports a variety of languages such as JavaScript, Ruby, Go, PHP, JAVA and more. It works with package.json as well as with Gemfiles.
```
fregante, line 1: It might be worth adding a full list of supported tools and languages at the bottom so that people searching for "npm" will find it

fregante:
> Alright, let's see if this needs further explanation before we spend time on this.

Quick installation instructions video we can point people to: https://user-images.githubusercontent.com/1402241/104827319-ed19fb00-5821-11eb-80f3-500f39d475b4.mov

safari/app-store-description.txt
```diff
@@ -0,0 +1,13 @@
...
+
+OctoLinker reads the source code you're currently viewing straight from the browser. 
+
+No source code will ever be sent to a server! Your source code stays in your browser. 
+
+Based on the file type, OctoLinker invokes a plugin which then executes a regular expression on the source code to find import statements.
+
+Relative file references are resolved through the GitHub tree API in your browser. External dependencies are resolved using the OctoLinker API (http://github.com/octoLinker/api).
+
+Only the name of a dependency along with the registry type is sent to this service.
```
stefanbuck, line 13:
```suggestion
Only the name of a dependency along with the registry type is sent to this service.


Supported tools and languages

composer, CSS, Deno, docker, dotnet, go, Haskell, Homebrew, HTML, Java, JavaScript, less, nodejs, npm, Paket, PHP, python, Rlang, Ruby, RubyGems, Rust, Sass, TypeScript, Vim, .Net
```

stefanbuck, line 13:
```suggestion
Only the name of a dependency along with the registry type is sent to this service.

Supported tools and languages:

composer, CSS, Deno, docker, dotnet, go, Haskell, Homebrew, HTML, Java, JavaScript, less, nodejs, npm, Paket, PHP, python, Rlang, Ruby, RubyGems, Rust, Sass, TypeScript, Vim, .Net
```
```

File: safari/app-store-keywords.txt [hunk 4]
```diff
@@ -0,0 +1 @@
+github,gh,git,safari,extension,browser,npm,python,issue,pull,request
```
stefanbuck, line 1: Do you know if there is a limit for keywords? I wonder whether or not we should list all tools and languages OctoLinker supports, like done in the description

fregante: It's 100 characters [apparently](https://www.google.com/search?client=safari&rls=en&q=app+store+keyword+limit&ie=UTF-8&oe=UTF-8), the UI doesn't say anything

stefanbuck:
> One of the photos is x801 instead of x800

I updated the store assets. Updating the video takes a bit more effort, mainly because I don't have to original video anymore.

safari/app-store-keywords.txt
stefanbuck, line 1:
```suggestion
github,gh,git,safari,extension,browser,npm,pull,request,javascript,php,npm,composer,go,ruby,python
```

stefanbuck: 98 characters 😉

PR #1182, opened 2021-01-16
Drop unnecessary details from manifest.json

fregante:
They are the defaults already https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_scripts


stefanbuck:
Good to know that those values are defaults now.

PR #1189, opened 2021-01-19
Update Safari version when using npm version script

stefanbuck:
Update Safari version when running `npm version` as outlined in https://github.com/OctoLinker/OctoLinker/pull/1167#discussion_r553619566

### DIFF
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 
-echo "Please summarize this release in one sentence"
-
-read note
-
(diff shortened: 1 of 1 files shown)


scripts/version.sh
```diff
@@ -1,10 +1,11 @@
...
-echo "Please summarize this release in one sentence"
-
-read note
-
+sed -E "s/MARKETING_VERSION = [0-9.]+/MARKETING_VERSION = $npm_package_version/" safari/Shared.xcconfig | tee safari/Shared.xcconfig
 dot-json assets/manifest.json version $npm_package_version
 dot-json packages/core/package.json version $npm_package_version
+
+echo "Please summarize this release in one sentence"
+read note
```
stefanbuck, line 8: The `read` command seem to interfere with `tee` command. Rearranging the command order is fixing this problem.

PR #1270, opened 2021-04-15
Remove php feature flag

xt0rted:
### Checklist:

### DIFF
--- a/packages/helper-settings/index.js
+++ b/packages/helper-settings/index.js
@@ -9,33 +9,11 @@ const defaults = {
     since: Date.now(),
     counter: 0,
   },
-  enableBetterPHP: false,
 };
-
-function isTestEnv() {
-  if (process.env.NODE_ENV === 'test') {
-    return true;
-  }
-
-  if (
-    typeof window !== 'undefined' &&
-    /OctoLinker\/OctoLinker\/blob\/.+\/e2e\/fixtures/.test(
-      window.location.pathname,
-    )
-  ) {
-    return true;
-  }
-
-  return false;
-}
 
 export const get = (key) => {
   if (process.env.OCTOLINKER_LIVE_DEMO) {
     return;
-  }
-
-  if (isTestEnv() && key === 'enableBetterPHP') {
-    return true;
   }
 
   return store[key];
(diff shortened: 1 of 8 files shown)


packages/helper-settings/index.js
```diff
@@ -9,35 +9,13 @@ const defaults = {
...
-    typeof window !== 'undefined' &&
-    /OctoLinker\/OctoLinker\/blob\/.+\/e2e\/fixtures/.test(
-      window.location.pathname,
-    )
-  ) {
-    return true;
-  }
-
-  return false;
-}
```
xt0rted, line 30: I wasn't sure if this should be left since it's not used anymore.

packages/resolver-hoogle-search/index.js
```diff
@@ -3,7 +3,7 @@ function handleResponse(hoogleSearchUrl, target, response) {
     return item.url.includes(`docs/${target.replace(/\./g, '-')}.html`);
   };
 
-  if (response.length > 0) {
+  if (response?.length > 0) {
```
xt0rted, line 6: This was causing an unresolved promise error in the `calls the hoogle search api` test since it wasn't setting the json value to anything.

packages/resolver-hoogle-search/__tests__/index.js
```diff
@@ -18,8 +18,8 @@ describe('hoogle-search', () => {
     assert.deepEqual(typeof hoogleSearch({ target }), 'function');
   });
 
-  it('calls the hoogle search api', () => {
-    hoogleSearch({ target })();
+  it('calls the hoogle search api', async () => {
+    await hoogleSearch({ target })();
```
xt0rted, line 22: `hoogleSearch` returns a promise but since this test isn't setting `response = []` (fixed in the other file) it was crashing with no real information, by awaiting it jest reports the failure correctly.

packages/helper-settings/index.js
stefanbuck: I think that's fine. I thought about using feature flags for other things as well, but given that testing isn't that straight forward we shouldn't continue with this approach

packages/plugin-php/index.js
```diff
@@ -1,64 +1,57 @@
...
+        )}.php`,
+      });
     }
 
     if (target.includes('\\')) {
       return [];
     }
 
     if (regExp === PHP_FUNC) {
-      return liveResolverQuery({
```
stefanbuck, line 52: For some reason, the ping through the OctoLinker API doesn't work for php.net anymore. A curl return 200, but a HEAD request made from nodejs fails with 503 `Service Temporarily Unavailable`. However, it should be fine to assume that the url we construct here are working in most cases like we do for the node.js core lib

PR #1292, opened 2021-05-08
Cargo: Support TOML inline table and cargo dev-dependencies

chawyehsu:
### Checklist:

### DIFF
--- a/packages/plugin-rust-cargo/index.js
+++ b/packages/plugin-rust-cargo/index.js
@@ -11,8 +11,10 @@ function linkDependency(blob, key, value) {
 function processTOML(blob, plugin) {
   let results = [];
   const json = TOML.parse(blob.toString());
+  const entries = Object.assign(json['dependencies'], json['dev-dependencies']);
 
-  Object.entries(json.dependencies).forEach(([name, version]) => {
+  Object.entries(entries).forEach(([name, value]) => {
+    const version = typeof value === "string" ? value : value.version;
     const linked = linkDependency.call(plugin, blob, name, version);
 
     results = results.concat(linked);
--- a/e2e/fixtures/rust/Cargo.toml
+++ b/e2e/fixtures/rust/Cargo.toml
@@ -20,3 +20,16 @@ env_logger = "0.5"
 
 # @OctoLinkerResolve(https://github.com/tokio-rs/bytes)
 bytes = "0.5"
+
+# @OctoLinkerResolve(https://github.com/tokio-rs/slab)
+slab = { version = "0.4.1", optional = true } # Backs `DelayQueue`
+
+[dev-dependencies]
+# @OctoLinkerResolve(https://github.com/tokio-rs/tokio)
+tokio = { version = "1.0", features = ["full"] }
+
+# @OctoLinkerResolve(https://github.com/tokio-rs/tokio)
+tokio-stream = { version = "0.1", path = "../tokio-stream" }
+
+# @OctoLinkerResolve(https://github.com/async-rs/async-std)
+async-std = "1.4"


packages/plugin-rust-cargo/index.js
```diff
@@ -11,8 +11,10 @@ function linkDependency(blob, key, value) {
 function processTOML(blob, plugin) {
   let results = [];
   const json = TOML.parse(blob.toString());
+  const entries = Object.assign(json['dependencies'], json['dev-dependencies']);
 
-  Object.entries(json.dependencies).forEach(([name, version]) => {
+  Object.entries(entries).forEach(([name, value]) => {
+    const version = typeof value === "string" ? value : value.version;
```
xt0rted, line 17: The regex being used (`tomlRegExKeyValue`) doesn't seem to work with object values like this. I had to add another regex that handled object values like this:

```js
export function tomlRegExObjectKeyValue(key, value) {
  const regexKey = escapeRegexString(key);
  const regexValue = escapeRegexString(value);

  return new RegExp(`(${regexKey})\\s*=\\s*{.*=\\s*"(${regexValue})"`, 'g');
}
```

And then call this based on if `value` is a string or object.

This is the full change:

```diff
@ -1,21 +1,27 @@
 import * as TOML from '@iarna/toml';
 import insertLink from '@octolinker/helper-insert-link';
-import { tomlRegExKeyValue } from '@octolinker/helper-regex-builder';
+import {
+  tomlRegExKeyValue,
+  tomlRegExObjectKeyValue,
+} from '@octolinker/helper-regex-builder';
import liveResolverQuery from '@octolinker/resolver-live-query';

-function linkDependency(blob, key, value) {
-  const regex = tomlRegExKeyValue(key, value);
+function linkDependency(blob, key, value, isStringMatch) {
+  const regex = isStringMatch
+    ? tomlRegExKeyValue(key, value)
+    : tomlRegExObjectKeyValue(key, value);
  return insertLink(blob, regex, this);
}

function processTOML(blob, plugin) {
  let results = [];
  const json = TOML.parse(blob.toString());
  const entries = Object.assign(json['dependencies'], json['dev-dependencies']);

  Object.entries(entries).forEach(([name, value]) => {
-    const version = typeof value === "string" ? value : value.version;
-    const linked = linkDependency.call(plugin, blob, name, version);
+    const isString = typeof value === 'string';
+    const version = isString ? value : value.version;
+    const linked = linkDependency.call(plugin, blob, name, version, isString);

    results = results.concat(linked);
  });
@ @@
```

@stefanbuck thoughts on this?

chawyehsu: How about passing the `isStringMatch` to `tomlRegExKeyValue` instead of creating a new regex function.

```js
export function tomlRegExKeyValue(key, value, isString) {
  const regexKey = escapeRegexString(key);
  const regexValue = escapeRegexString(value);
  const reStr = isString ? `(${regexKey})\\s*=\\s*"(${regexValue})"`
    : `(${regexKey})\\s*=\\s*{.*=\\s*"(${regexValue})"`;

  return new RegExp(reStr, 'g');
}
```

```js
function linkDependency(blob, key, value, isString) {
  const regex = tomlRegExKeyValue(key, value, isString);
  return insertLink(blob, regex, this);
}

function processTOML(blob, plugin) {
  let results = [];
  const json = TOML.parse(blob.toString());
  const entries = Object.assign(json['dependencies'], json['dev-dependencies']);

  Object.entries(entries).forEach(([name, value]) => {
    const isString = typeof value === "string";
    const version = isString ? value : value.version;
    const linked = linkDependency.call(plugin, blob, name, version, isString);

    results = results.concat(linked);
  });

  return results;
}
```

stefanbuck: Sorry for the late response. I think both are valid approaches and at the end it doesn't really matter. I don't see this toml regex being used in other places than `plugin-rust-cargo/index.js`. Unless @xt0rted disagrees I think we can go ahead with the single function suggestions from @chawyehsu

chawyehsu:
I'm planning to make rust's/cargo's support more comprehensive, including `std` and relative modules. The question is I don't quite understand how to handle relative modules, and dependencies without specifying the version.

**std**: https://doc.rust-lang.org/std/
```rust
use std::{fs, io};
use std::path::Path;
```

**relative mods**
1. in Cargo.toml
`path = "../tokio-stream"` to use relative crate as dependencey, documented in https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies
```toml
tokio-stream = { version = "0.1", path = "../tokio-stream" }
```
2. in .rs source file
`crate::` to import mods from the current crate root
```
use crate::io;
use crate::path::Path;
```

**no version number**
```toml
[dependencies]
rand = { git = "https://github.com/rust-lang-nursery/rand" }
```

packages/plugin-rust-cargo/index.js
chawyehsu: I prefer the single function, thoughts? @xt0rted

xt0rted: I prefer two functions since it's clearer what's goin on, but either way is fine with me.

chawyehsu:
I created a repo for testing the linker's effect on rust/cargo: https://github.com/chawyehsu/octolinker-rust-test-cases

PR #1374, opened 2021-10-10
Remove unused dependency sinon

stefanbuck:
I just noticed that sinon is not used anymore so should be save to remove


stefanbuck:
Upps thanks for the hint

PR #1414, opened 2021-12-10
Bind window object to requestIdleCallback assignment

xt0rted:
### Checklist:

The issue reported in https://github.com/OctoLinker/OctoLinker/issues/942#issuecomment-976689694 is a result of how Firefox is treating the `window.requestIdleCallback` assignment. In Chrome this works fine, but in Firefox any call to it results in the error:

>TypeError: 'requestIdleCallback' called on an object that does not implement interface Window.

I had the same results in v94 and v96.

Another option is to remove the conditional check and [shim the function](https://developers.google.com/web/updates/2015/08/using-requestidlecallback#checking_for_requestidlecallback) at the top of the file with:

```js
window.requestIdleCallback =
  window.requestIdleCallback ||
  function (cb) {
    const start = Date.now();
    return setTimeout(() => {
      cb({
        didTimeout: false,
        timeRemaining: () => Math.max(0, 50 - (Date.now() - start)),
      });
    }, 1);
  };
```

Could you test these in Safari? I don't have my MacBook setup right now so I can't confirm if it still work with that.

### DIFF
--- a/packages/core/app.js
+++ b/packages/core/app.js
@@ -50,7 +50,7 @@ async function run(rootElement) {
     });
 
   // Prefetch live resolver results in background
-  (window.requestIdleCallback || setTimeout)(() => {
+  (window.requestIdleCallback || setTimeout).bind(window)(() => {
     loader(matches);
   });
 }


xt0rted:
@stefanbuck the tests started failing, I think due to GitHub's html changing.

stefanbuck:
> @stefanbuck the tests started failing, I think due to GitHub's html changing.

Looks like. I thought we run `./packages/blob-reader/scripts/update-fixtures.sh` to ensure we have the latest fixture version but this is not true. After running the script a bunch of blob reader tests started failing. Most of them related to the diff view. I'll look into this

stefanbuck:
It took me a while to figure out which part changed but here we go https://github.com/OctoLinker/OctoLinker/pull/1427

nikoladev:
I can confirm that this change fixed OctoLinker  for me in Firefox on macOS. After hacking the change into the `.xpi` manually OctoLinker started working again.

stefanbuck:
Thanks for help with this @nikoladev. I'm going to publish a new version later today.

stefanbuck:
Fixed in v6.10.1 which is already available

jdreesen:
It's working again in my Firefox (on Ubuntu), thank you so much! :heart:

PR #1420, opened 2021-12-15
Drop jQuery

fregante:
- Replaces https://github.com/OctoLinker/OctoLinker/pull/1049

Untested, basically a manual search and replace:

- if a single element was expected, I used `querySelector` + `?.`
- if multiple elements were expected, I used `querySelectorAll` in a wrapper that produces an array
- use $() and $$() match the console utilities: https://developer.chrome.com/docs/devtools/console/utilities/#querySelector-function


stefanbuck:
This is great 🎉 I'll take a closer look at your changes over the holidays. Just one note on the failing test. Sometime `jsdom` is not behaving as expect and I think for some reason `querySelectorAll` is not available. I can dig into this as part of the review.

fregante:
Ok it seems to work now, but I probably messed something up because the tests are still failing.

xt0rted:
@fregante try rebasing now that the tests & dependencies were updated.

packages/blob-reader/helper.js
```diff
@@ -1,19 +1,24 @@
...
-function getBlobWrapper(rootElement = document) {
-  const ret = [
-    ...[].slice.call(rootElement.getElementsByClassName('blob-wrapper')),
-    ...[].slice.call(rootElement.getElementsByClassName('js-blob-wrapper')),
-    ...[].slice.call(
-      rootElement.querySelectorAll('[class*="highlight-source-"]'),
-    ),
-  ];
+function getBlobCodeInner(el) {
+  return $$('.blob-code-inner');
```
xt0rted, line 10: Should `el` be passed to `$$()`?

packages/blob-reader/helper.js
```diff
@@ -102,11 +105,16 @@ function getLineNumber(el) {
   }
 
   // split diff view
-  let lineNumber = $(el).closest('td').prev().data('line-number');
+  let lineNumber = el
+    .closest('td')
+    ?.previousElementSibling?.getAttribute('line-number');
```
xt0rted, line 110:
```suggestion
    ?.previousElementSibling?.getAttribute('data-line-number');
```

packages/blob-reader/helper.js
```diff
@@ -102,11 +105,16 @@ function getLineNumber(el) {
...
 
   // unified diff view
   if (!lineNumber) {
-    lineNumber = $(el).closest('tr').find('td').data('line-number');
+    lineNumber = el
+      .closest('tr')
+      ?.querySelector('td')
+      ?.getAttribute('line-number');
```
xt0rted, line 117:
```suggestion
      ?.getAttribute('data-line-number');
```

xt0rted (changes requested):
I was able to track down why this isn't working. I left comments on the spots I could but there's one more spot I can't since it wasn't edited as part of this PR.

https://github.com/OctoLinker/OctoLinker/blob/a93670e0922a0a96a8224622366ba44ff152de9e/packages/blob-reader/helper.js#L120-L124

This needs to change to:

```js
if (lineNumber) {
  if (!Number.isNaN(lineNumber)) {
    return parseInt(lineNumber, 10);
  }
}
```

jQuery was returning the data attribute values as numbers, but `getAttribute()` returns a string and there's code downstream that needs them as a number.

packages/blob-reader/helper.js
fregante: The perils of _not_ using every lint rule available 😃

fregante:
Boom, tests pass

stefanbuck:
I did a couple of manual spot checks and seems to work. Really great work from both of your.

PR #1427, opened 2021-12-29
Fix blob reader

stefanbuck:
GitHub made some changes to the DOM which caused OctoLinker to not function well in some scenarios like in a diff view. See comments for full disclosure

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
         node-version: 16
         cache: 'yarn'
     - run: yarn install
+    - run: ./packages/blob-reader/scripts/update-fixtures.sh
     - run: yarn test
     - run: yarn build
     - name: Run E2E
--- a/packages/blob-reader/helper.js
+++ b/packages/blob-reader/helper.js
@@ -74,7 +74,7 @@ function getPath(el) {
   }
 
   // When current page is a gist, get path from blob name
-  if (!ret && isGist()) {
+  if (isGist()) {
     ret = $('.gist-blob-name', el.parentElement).text().trim();
     if (ret && !ret.startsWith('/')) {
       ret = `/${ret}`;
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ jobs:
         node-version: 16
         cache: 'yarn'
     - run: yarn install
+    - run: ./packages/blob-reader/scripts/update-fixtures.sh
```
stefanbuck, line 19: For some reason I thought we run this script on CI already to ensure our unit tests run against the last dom, but turns out this wasn't the case. I was considering to add this as a `pretest` npm script, but then every local test run will update the fixtures as well which might be confusing to other contributors so I went with this approach.

packages/blob-reader/helper.js
```diff
@@ -137,7 +122,7 @@ function diffMetaInformation(el) {
   const td = $(el).closest('td').get(0);
 
   // Blob view
-  if (td.classList.contains('js-file-line')) {
+  if (td.id.startsWith('LC') || isGist()) {
```
stefanbuck, line 125: This was the major change that GitHub did which caused most of the blob reader unit tests to fail. Previously `js-file-line` was only being used when looking at a single file, but now this class is always present even in diffs. The new approach checks the `id` attribute which starts with `LC` for single files and `diff-` in a diff scenario. Overall this exercise reminded me how complicated OctoLinker got over the past years.

packages/blob-reader/helper.js
```diff
@@ -74,34 +74,19 @@ function getPath(el) {
...
-      filePath = $('a', $fileHeader).text();
-      repoPath = $('a', $fileHeader).attr('href');
-    }
-
-    if (repoPath && filePath) {
-      ret = mergeRepoAndFilePath(repoPath, filePath);
-    }
+    const filePath = $('a', $fileHeader).text();
+    const repoPath = $('a', $fileHeader).attr('href');
+    ret = mergeRepoAndFilePath(repoPath, filePath);
```
stefanbuck, line 89: This was another change to the dom that caused code snippets part of a review to be broken because OctoLinker was unable to construct the file path which is being used to invoke the related plugin for this file extension.

.github/workflows/ci.yml
xt0rted: Would it be good to run this on a schedule, like once a week or so, and automatically PR the changes to keep the files in sync?

stefanbuck: Good idea. Is this something you want to look into?

xt0rted: We were doing something very similar to this for [GitHub-Dark](https://github.com/StylishThemes/GitHub-Dark/blob/master/.github/workflows/regenerate.yml). A PR created using the `GITHUB_TOKEN` won't trigger workflows so a PAT or GitHub App will be needed, similar to the Dependabot auto-merge workflow.

PR #1431, opened 2022-01-03
Solidity: improve path matching

lucaperret:
This PR improve the matching of Solidity files by using the JS regex.
Now matching also:
```solidity
import {YourContract} from "./import.sol";
```


stefanbuck:
Thanks for the fix. Looks like a few golang related end to end tests started failing. Unfortunately I'm unable to take a look before mid next week. Really sorry for the inconveniences, but I'll get back asap

stefanbuck:
All end 2 end tests are passing after a re-run. Due to their nature of testing the entire system including downstream services they sometimes fail unexpectedly.

stefanbuck:
After some housekeeping tasks we're going to release a new version. Thanks again for your contribution @lucaperret

PR #1467, opened 2022-01-25
Support built-in packages with `node:` prefix

### LINKED ISSUE #1466: Node built-in modules aren't detected when prefixed with `node:`
### Expected Behavior

Since v16 (and in previous versions through backporting) Node.js supports an alternative way to import built-in modules (like `fs`) which involves prefixing the module's name with `node:`. Using this syntax, importing `fs` looks like this:

```js
import fs from 'node:fs'
```

This is functionally equivalent to the following:

```js
import fs from 'fs'
```

Hence, it should also link to the same thing: the Node.js documentation page for that module.

### What actually happened?

Currently, Octolinker generates a link with the `href` attribute set to `node:process`, which results in Chrome trying to open the "URL" `process` with an application assigned to the `node` scheme/protocol. That doesn't seem right.

### URL

https://github.com/sindresorhus/globby/blob/43eccf41a800a5dedb3b3fcdf313f7daf5c199ff/index.js#L1

### Anything else we should know?

_No response_

xt0rted:
### Checklist:

Looks to be supported in:
- `v16.0.0`, `v14.18.0` (ESM import and CommonJS require())
- `v14.13.1`, `v12.20.0` (only ESM import)
- In TypeScript with the latest versions of `@types/node`

Fixes #1466

### DIFF
--- a/packages/plugin-javascript/index.js
+++ b/packages/plugin-javascript/index.js
@@ -74,6 +74,10 @@ export default {
   name: 'JavaScript',
 
   resolve(path, [target]) {
+    if (target.startsWith('node:')) {
+      target = target.substring(5);
+    }
+
     if (isURLImport(target)) {
       return resolverTrustedUrl({ target });
     }
--- a/e2e/fixtures/javascript/index.js
+++ b/e2e/fixtures/javascript/index.js
@@ -1,6 +1,9 @@
 // @OctoLinkerResolve(https://nodejs.org/api/fs.html)
 require('fs');
 
+// @OctoLinkerResolve(https://nodejs.org/api/fs.html)
+require('node:fs');
+
 // @OctoLinkerResolve(https://github.com/lodash/lodash)
 require("lodash");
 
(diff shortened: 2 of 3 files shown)


packages/plugin-javascript/index.js
```diff
@@ -74,6 +74,10 @@ export default {
   name: 'JavaScript',
 
   resolve(path, [target]) {
+    if (target.startsWith('node:')) {
+      target = target.substring(5);
```
stefanbuck, line 78: Today I learned about `substring`. Nice one

xt0rted: I came across this after creating the PR. Not sure if there's a preferred way of doing this.

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-slice.md

stefanbuck: Personal preference I would say. `substring` seems to be less well known, but don't worry about it.

PR #1495, opened 2022-02-20
Link github actions to their pinned version

xt0rted:
### Checklist:

This adjusts GitHub Actions links so if they include a tag/sha we link directly to that instead of the default branch. I figured since most actions are pinned to a version it'd make more sense to link to that instead of the default branch. I'm not sure if we want to do this or not, I'm fine with it either way.


stefanbuck (approved):
Great Idea

PR #1521, opened 2022-03-18
Improve Safari app icons

fregante:
I always felt like my earlier alignment wasn't working as well as I hoped.

Before | After
--- | ---
 |


fregante:
Also at some point I hope to port the extension to iOS, if this goes well:

- https://github.com/npmhub/npmhub/pull/152

Also related:

- https://github.com/refined-github/refined-github/pull/4787

PR #1536, opened 2022-04-05
Update resolve url

xt0rted:
The change in url caused the e2e tests to start to fail.

The other urls are technically wrong because `https://golang.org/pkg/context` redirects to `https://pkg.go.dev/context` but updating them causes the tests to fail. This is probably because of https://github.com/OctoLinker/OctoLinker/blob/4a54d31577f21659e33d0fa1cdfff00dc9d02ad3/packages/plugin-go/index.js#L53

### Checklist:

### DIFF
--- a/e2e/fixtures/go/foo.go
+++ b/e2e/fixtures/go/foo.go
@@ -1,12 +1,12 @@
 package master
 
-// @OctoLinkerResolve(https://golang.org/pkg/context)
+// @OctoLinkerResolve(https://pkg.go.dev/context)
 import "context"
 
-// @OctoLinkerResolve(https://golang.org/pkg/math/big)
+// @OctoLinkerResolve(https://pkg.go.dev/math/big)
 import "math/big"
 
-// @OctoLinkerResolve(https://golang.org/pkg/k8s.io/apimachinery/pkg/api/errors)
+// @OctoLinkerResolve(https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors)
 import "k8s.io/apimachinery/pkg/api/errors"
 
 // @OctoLinkerResolve(https://github.com/kubernetes/api/tree/master/core/v1)
--- a/e2e/fixtures/go/foo.go
+++ b/e2e/fixtures/go/foo.go
@@ -9,7 +9,7 @@ import "math/big"
 // @OctoLinkerResolve(https://golang.org/pkg/k8s.io/apimachinery/pkg/api/errors)
 import "k8s.io/apimachinery/pkg/api/errors"
 
-// @OctoLinkerResolve(https://golang.org/pkg/k8s.io/api/core/v1)
+// @OctoLinkerResolve(https://github.com/kubernetes/api/tree/master/core/v1)
 import corev1 "k8s.io/api/core/v1"
 
 // @OctoLinkerResolve(https://github.com/ethereum/go-ethereum/tree/master/common)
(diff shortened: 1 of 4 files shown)


stefanbuck:
Good catch. Yes, I think we should update the base url. The constructed url is sent to the server which does a ping (HTTP HEAD request) to ensure the url is reachable. Therefore updating the base url here will be enough. Do you want to make this change part of this pull request?

stefanbuck (approved):

stefanbuck:
> @stefanbuck does this list need to be updated?
> 
> https://github.com/OctoLinker/OctoLinker/blob/4a54d31577f21659e33d0fa1cdfff00dc9d02ad3/packages/helper-grammar-regex-collection/go.js#L6-L14

Yes I think so. I'm glad you spotted it

e2e/fixtures/go/foo.go
```diff
@@ -1,15 +1,15 @@
...
-// @OctoLinkerResolve(https://golang.org/pkg/math/big)
+// @OctoLinkerResolve(https://pkg.go.dev/math/big)
 import "math/big"
 
-// @OctoLinkerResolve(https://golang.org/pkg/k8s.io/apimachinery/pkg/api/errors)
+// @OctoLinkerResolve(https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors)
 import "k8s.io/apimachinery/pkg/api/errors"
 
-// @OctoLinkerResolve(https://golang.org/pkg/k8s.io/api/core/v1)
+// @OctoLinkerResolve(https://pkg.go.dev/k8s.io/api/core/v1)
```
xt0rted, line 12: @stefanbuck my initial change got the tests passing but now they're failing again. This url resolved to `https://github.com/kubernetes/api/tree/master/core/v1` yesterday but now it's resolving to `https://golang.org/pkg/k8s.io/api/core/v1` again. The tests won't pass with either url though. Any ideas what's going on here? Is the api caching the wrong url?

stefanbuck: > Is the api caching the wrong url?

Yes, I think this is what is happening here. In https://github.com/OctoLinker/api/pull/426/files#diff-90e29dc49ee3814a005e7dd6d34a3bd3bacbcc1daba54f195e21cbb70bca564aR27 I updated the caching of 404 ping results from 15min to 48 hours. Not sure how to resolve this issue best to be honest.

stefanbuck: btw for me it still resolves to `https://github.com/kubernetes/api/tree/master/core/v1`. The reason is because my location hits another Redis cache (we have a Redis cache running in four major regions). This makes it even more complicated. Maybe we should just merge it and get the new release out

xt0rted: I just removed the revert I tried so the test now points to the github url, which is also what I get when testing with one of the api PR instances.

Does this change need any tests? https://github.com/OctoLinker/OctoLinker/pull/1536/files#diff-f1c4ee55132845e20cf1dd168feb8f34ed36e11d1651ecb94f107c059d7b3fa0

stefanbuck: fyi https://github.com/OctoLinker/api/pull/439

xt0rted: The tests passed, better merge this before they start to fail again 😄

stefanbuck: Maybe GA are running in different regions, but yeah let's merge this

PR #1537, opened 2022-04-05
Add concurrency settings

xt0rted:
This will cancel an in-progress workflow if another instance of it starts.

https://docs.github.com/en/actions/using-jobs/using-concurrency

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,9 @@ on:
     branches:
       - main
   pull_request:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


stefanbuck (approved):
This is a really good idea to reduce the load on the API while running the tests also it will make the feedback loop faster in case a new commit gets added. Out of curiosity, was the latter the reason why you made this change?

xt0rted:
I don't think it'll speed up the feedback loop because a new workflow starts for each commit and they can run in parallel. In most cases you don't care about the results of a run except for the most recent commit so if we can stop them from running then it should help reduce load overall, and maybe in the api too.

These settings are more useful in private repos since you pay for build minutes and artifact storage, but it seems useful in public repos too.

PR #1538, opened 2022-04-06
Remove update notification

stefanbuck:
This is something I wanted to do since quite some time. There was user feedback that indicated that this message is not really useful and from my own experience I can tell that I pay little attention to such release notifications as well


xt0rted:
>✕ resolves import corev1 "k8s.io/api/core/v1" to https://github.com/kubernetes/api/tree/master/core/v1 (5000 ms)

😡

PR #1540, opened 2022-04-06
New release flow

stefanbuck:
A discussed with @xt0rted this updates thee release flow. Now it is just a matter of creating a release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,23 +1,53 @@
 name: Release
 on:
-  push:
-    tags:
-      - v*
+  release:
+   types: [published]
 jobs:
   release:
     runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
     steps:
-    - uses: actions/checkout@v3
-    - uses: actions/setup-node@v3
-      with:
-        node-version: 16
-    - run: yarn install
-    - run: yarn build
-    - run: 'yarn release'
-      env:
-        CLIENT_ID: ${{ secrets.PUBLISHER_CHROME_CLIENT_ID }}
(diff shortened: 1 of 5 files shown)


assets/manifest.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "OctoLinker",
-  "version": "6.10.2",
+  "version": "0.0.0",
```
stefanbuck, line 3: From now on, we will not store the actual version number in git anymore.

.github/workflows/release.yml
```diff
@@ -1,23 +1,53 @@
...
+
+      - name: Checkout repository
+        uses: actions/checkout@v3.0.0
+
+      - name: Install node
+        uses: actions/setup-node@v3.1.0
+        with:
+          node-version: 16
+
+      - name: Get changelog entry
```
xt0rted, line 26: We don't have a change log, so this step isn't needed unless we want to start doing that. If we do it has to be in this format for the action to work (most of the repos listed in my profile readme use this format since I'm also using this action with them) https://keepachangelog.com

stefanbuck: Thanks, I'll remove it for now.

.github/workflows/release.yml
```diff
@@ -1,23 +1,47 @@
...
+      - run: yarn version ${{ steps.tag_name.outputs.current_version }}
+      - run: yarn build
+
+      - name: Package dist folder
+        run: tar -zcvf octolinker-${{ steps.tag_name.outputs.current_version }}.tar.gz ./dist
+
+      - name: Upload release to GitHub
+        uses: softprops/action-gh-release@v1
+        with:
+          body: ${{ steps.changelog_reader.outputs.changes }}
```
xt0rted, line 36:
```suggestion
```

This isn't needed since the changelog step was removed.

PR #1696, opened 2022-11-13
Prevent GitHub's code navigation styling on our links

xt0rted:
This resolves the issue mentioned in https://github.com/OctoLinker/OctoLinker/issues/1695#issuecomment-1312817951 which should be backwards compatible with the old site. I'm not sure if there's a better way to do this, but based on what their code's doing this attribute is what stops the `pl-token` class from being added which is where the hover style comes from. They set it to `true` so I figured `false` made more sense in this context.

### Checklist:


stefanbuck (approved):
I couldn't figure out a better way either 😄

PR #1697, opened 2022-11-13
Support action workflow run view

xt0rted:
Add support for linking action workflows in the [new actions UI](https://github.blog/2022-10-20-improving-navigation-for-github-actions/). This is more of a proof of concept, but figured I'd PR it for feedback. Unfortunately there's no other classes or attributes I can see that would let us get this link like we do on the source pages.

Example: https://github.com/OctoLinker/OctoLinker/actions/runs/3457001003/workflow

### Checklist:

### DIFF
--- a/packages/blob-reader/helper.js
+++ b/packages/blob-reader/helper.js
@@ -86,6 +86,13 @@ function getPath(el) {
     ret = $('.js-permalink-shortcut')?.getAttribute('href');
   }
 
+  // When current page is an actions run workflow file
+  if (!ret) {
+    ret = $(
+      '.actions-fullwidth-module .text-small .Link--secondary',
+    )?.getAttribute('href');
+  }
+
   // When current page is a gist, get path from blob name
   if (isGist()) {
     ret = $('.gist-blob-name', el.parentElement)?.textContent.trim();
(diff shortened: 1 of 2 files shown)


packages/blob-reader/helper.js
```diff
@@ -86,6 +86,13 @@ function getPath(el) {
     ret = $('.js-permalink-shortcut')?.getAttribute('href');
   }
 
+  // When current page is an actions run workflow file
+  if (!ret) {
+    ret = $(
+      '.actions-fullwidth-module .text-small .Link--secondary',
```
stefanbuck, line 92: Maybe we should switch to https://github.com/refined-github/github-url-detection 

Jumpcat.dev is using `github-url-detection` and it works really well.

xt0rted: How would that help get the url from this link? The value of it is `https://github.com/OctoLinker/OctoLinker/blob/e9e60b630a37e980e48a816e4aaf6355a3fd8707/.github/workflows/ci.yml` and it's the only spot on the page with it.

image

stefanbuck: I shouldn't comment on PRs after midnight 🙈

PR #1698, opened 2022-11-13
Add support for linking to vscode extensions

xt0rted:
While setting up a project recently I found myself having to link to the extensions I was adding in the PR so it was easier to review and figured it might be nice if we linked to them too.

From what I've found there's no easy API we can query to try to get the repo link for these which is kind of annoying. Here's all I was able to find on it https://github.com/microsoft/vsmarketplace/issues/238.

Some example files I used while adding this in:

- https://github.com/actions/go-dependency-submission/blob/main/.devcontainer/devcontainer.json
- https://github.com/actions/runner/blob/main/.devcontainer/devcontainer.json
- https://github.com/microsoft/TypeScript/blob/main/.vscode/extensions.json

### Checklist:


stefanbuck (approved):
Yeah would be nice if we could link to the repo, but we do the same for other things liker docker images so shouldn't be a big concern.

xt0rted:
@stefanbuck the test failure looks to be related to #1702. I was getting similar errors in my projects before I fully moved to modules.

PR #1716, opened 2022-11-22
Add explicit commit message prefix

xt0rted:
I talked to support about the commit message prefixes that started being used here (happened in one of my repos too) and they said the way to prevent it to by setting the `commit-message` property in the dependabot config.

### Checklist:


stefanbuck (approved):
Good to know. Thanks for looking into this

xt0rted:
I forgot about this PR. I could have merged this instead of creating #1724 😆

